PR #23, opened 2016-08-19
Adding documentation for receiving updates

### LINKED ISSUE #13: Getting updates [receving-updates.md]

Poincare:
Will eventually fix #13.

### DIFF
--- a/source/receiving-updates.md
+++ b/source/receiving-updates.md
@@ -3,6 +3,59 @@ title: Getting updates from the server
 order: 6
 ---
 
+Apollo Client caches the results of queries and then uses this cache in order to resolve parts of queries. However, what happens if the information in our cache goes out of date, i.e. the cache becomes stale? How do we make sure that we can update the cache with information if information changes on the server? How will our existing queries get updates to the information they've asked for? These are questsions that this section should answer.
+
+A momentarily stale cache is an unavoidable problem. There's no feasible way to have a client-side cache and make sure that the cache will *always* reflect exactly the information that is available on the server. For pretty much any application, this isn't too much of a problem: your UI may be slightly out-of-date temporarily, but, it'll sync soon enough. There are a few strategies to make sure that Apollo Client is eventually consistent with the information available to your server. These are: refetches, polling queries and GraphQL subscriptions.
+
 ## Refetches
+Refetches are the simplest way to force a portion of your cache to reflect the information available to your server. Essentially, a refetch forces a query to hit the server, bypassing the cache entirely. The result of this query, just like all other query results, will update the information available in the cache.
+
+For example, continuing with the GitHunt schema, we may have the following on a component:
+
+```javascript
+const FEED_QUERY = gql`
+    query Feed($type: FeedType!, $offset: Int, $limit: Int) {
+        feed($type: NEW, offset: $offset, lim) {
+            createdAt
+            commentCount
+            score
+            id
+            respository {
+                // etc.
+            }
+        }
+    }`;
+
+@graphql(FEED_QUERY);
+class FeedComponent extends Component {
+    // ...
+    someFunction() {
+        this.props.data.refetch();
+    }
+    // ...
+}
+```
+
+In particular, we have the method `this.props.refetch`, which allows us to refetch the query associated with the `FeedCompoment`. This means that instead of resolving information about the `feed` field from the cache (even if we have it!), the query will hit the server and will update the cache with new results from the server.
+
+So, if there's been some kind of update in the information that the query requests (e.g. a new repository added to the feed), the Apollo Client store will have the update and the UI will re-render as necessary.
+
+In order for refetches to be a viable strategy, you must have some idea as to when you should refetch the information (i.e. when information in the cache has gone stale). This is possible in many circumstances. For example, you could imagine refetching the whole feed when the client adds a new repository to it. But, there are cases in which this does not work, e.g. some *other* client decides to insert a repository into the GitHunt feed. Then, our client has no idea that this has happened and won't see the new feed item until the page is refreshed. One solution to that problem is polling.
+
 ## Polling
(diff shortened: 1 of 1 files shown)


Poincare:
Review please? @tmeasday @helfer

source/receiving-updates.md
```diff
@@ -3,6 +3,61 @@ title: Getting updates from the server
...
+class FeedComponent extends Component {
+  // ...
+  someFunction() {
+    this.props.data.refetch();
+  }
+  // ...
+}
+```
+
+In particular, we have the method `this.props.refetch`, which allows us to refetch the query associated with the `FeedCompoment`. This means that instead of resolving information about the `feed` field from the cache (even if we have it!), the query will hit the server and will update the cache with new results from the server.
```
stubailo: This.props.data.refetch

source/receiving-updates.md
```diff
@@ -3,6 +3,61 @@ title: Getting updates from the server
...
+
 ## Polling
+If you have a query whose result can change pretty frequently, it probably makes sense to consider making a polling query. A polling query is a GraphQL query which is fired on a particular interval and every time it is fired, it is fired as a refetch, i.e. no part of it is resolved from the cache. By doing this, the portion of the cache that the query touches will be updated on the polling interval and will be consistent with the information that is on the server.
+
+Continuing with our refetch example, we can add a polling interval with an additional option:
+
+```javascript
+@graphql(FEED_QUERY, {
+  options: (props) => {
```
stubailo: This can be a plain object. Doesn't need to be a thunk

stubailo: I'd add a suggestion that people shouldn't do polling on short intervals. I'd suggest 10 or 20 seconds because I feel like people should use subscriptions when they need fast updates and the docs should recommend doing so.

source/receiving-updates.md
```diff
@@ -3,6 +3,61 @@ title: Getting updates from the server
...
+      respository {
+        // etc.
+      }
+    }
+  }`;
+
+@graphql(FEED_QUERY);
+class FeedComponent extends Component {
+  // ...
+  someFunction() {
```
tmeasday: Can we provide a more concrete example? Is it too weird to have a "refresh" button and `onRefreshClicked`?

source/receiving-updates.md
```diff
@@ -3,6 +3,61 @@ title: Getting updates from the server
...
+    this.props.data.refetch();
+  }
+  // ...
+}
+```
+
+In particular, we have the method `this.props.refetch`, which allows us to refetch the query associated with the `FeedCompoment`. This means that instead of resolving information about the `feed` field from the cache (even if we have it!), the query will hit the server and will update the cache with new results from the server.
+
+So, if there's been some kind of update in the information that the query requests (e.g. a new repository added to the feed), the Apollo Client store will have the update and the UI will re-render as necessary.
```
tmeasday: I think we should probably have an example of this, because binding the `refetch` to the mutation completion callback in the container component (which IMO is the "right" way to do it) isn't necessarily clear to people. Probably we should do it in githunt too (how does the re-ordering of the feed when you vote work anyway?)

Alternatively we could have this in the patterns section maybe and link to it here.

I think the code should look something like:

```js
@graphql(FEED_QUERY);
@graphql(VOTE_MUTATION, {
  props({ mutate }) {
    const { refetch } = this.props.data;
    this.vote = (repoName) => {
       mutate({ repoName })
         then(refetch);
    }
  }
});
```

stubailo: > how does the re-ordering of the feed when you vote work anyway?

I don't think it makes sense to re-order when you click. This sounds like a cool demo but basically no real apps do this.

Poincare: I'm not sure if this is the "right" way since we now have a (pretty undocumented) `refetchQueries` option on the mutation, which specifies a list of queries to refetch once a mutation returns. That is probably the right way to refetch the a query on a mutation.

I'm not sure if `react-apollo` currently supports `refetchQueries` but if it does not, it definitely should.

source/receiving-updates.md
Poincare: Fixed.

source/receiving-updates.md
Poincare: Fixed.

source/receiving-updates.md
Poincare: Fixed, although I omitted the implementation of the refresh button. Instead, I added a text description of the button and changed the name of `someFunction` to `onRefreshClicked`.

Poincare:
Made some fixes according to the comments

source/receiving-updates.md
tmeasday: Oh, interesting point @Poincare, let's get that documented, I'll open an issue and we can chase it up.

tmeasday: Actually, thinking about it more @Poincare, you can pass all mutation arguments into the `mutate` function created by `react-apollo`. So I'm pretty sure you can just do

```
graphql(MUTATION, {
  props: ({ mutate }) ({
    doSomething(arg) {
      mutate({ x: arg }, { refetchQueries: 'y' });
    }
  })
});

Poincare: I think the "refetch on mutation" should be introduced with `fetchMore` and `updateQueries` because of two reasons:

1. This isn't what you should do most of the time.

Most of the time, as long as the mutation returns an id, your UI should be updated automatically.

2. `refetchQueries` is supposed to be a lazy way of doing `updateQueries`

Basically, if you don't want to write a function that manually incorporates the result of a mutation using `updateQueries`, you should use `refetchQueries`. It should not be seen as the "ideal way" of doing things.

stubailo: I dunno - I'd say maybe we should first introduce the "screw it, just refetch everything" and then teach people how to optimize the stuff later? At least, that's what I would do.

Poincare: I am fine with doing that on the page that describes `updateQueries` and `fetchMore` (i.e. place `refetchQueries` first on that page), but I don't think it should be presented on the same page as "Receiving updates for your data" since most of the time, you don't even need `refetchQueries` at all.

tmeasday:
Looks great merging!

PR #24, opened 2016-08-22
Fetch more docs

### LINKED ISSUE #14: `fetchMore` section of `cache-updates.md`

Poincare:
Documentation for `fetchMore`. Fixes #14.

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -3,7 +3,7 @@ title: Updating your UI
 order: 7
 ---
 
-Assigning object identifiers to your result objects using `dataIdFromObject` allows Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. In particular, if you want to add something to a list of objects or if there are some objects that you can't assign an object identifier to, you should use either `fetchMore` or `updateQueries` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
+Within Apollo Client, we can specify a function through the `dataIdFromObject` option in order to give object identifiers to objects that our queries return. These allow Apollo Client to figure out when two different results actually point to exactly the same object. These object identifiers also allow Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. In particular, if you want to add something to a list of objects or if there are some objects that you can't assign an object identifier to, you should use either `fetchMore` or `updateQueries` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
 
 ## `fetchMore`
 
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -3,9 +3,64 @@ title: Updating your UI
 order: 7
 ---
 
+Assigning object identifiers to your result objects using `dataIdFromObject` allows Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. In particular, if you want to add something to a list of objects or if there are some objects that you can't assign an object identifier to, you should use either `fetchMore` or `updateQueries` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
+
 ## `fetchMore`
 
-In order to incorporate the result of a query into the result of another query
+Typically, `fetchMore` is used to manually update the result of one query based on the data returned by another query. Most often, it is used to handle pagination. In our GitHunt example, we have a paginated feed that displays a list of GitHub respositories. When we hit the "Load More" button, we don't Apollo Client to throw away the repository information it has already loaded. Instead, it should just append the newly loaded repositories to the list of repositories that Apollo Client already has in the store. With this update, our UI component should re-render and show us all of the available repositories.
+
+This is possible with `fetchMore`. The `fetchMore` method allows us to fetch another query and incorporate that query's result into the result that our component query previously received. We can see it in action within the GitHunt code:
+
+```javascript
+const FeedWithData = graphql(FEED_QUERY, {
+  props({ data: { loading, feed, currentUser, fetchMore, variables } }) {
+    return {
+      loading,
+      feed,
+      currentUser,
+      fetchMore() {
+        return fetchMore({
+          variables: {
+            offset: feed.length,
+          },
+          updateQuery: (prev, { fetchMoreResult }) => {
+            if (!fetchMoreResult.data) { return prev; }
+            return Object.assign({}, prev, {
+              feed: [...prev.feed, ...fetchMoreResult.data.feed],
+            });
+          },
+        });
+      },
+    };
+  },
+})(Feed);
+```
+
+We have two components here: `FeedWithData` and `Feed`. The `FeedWithData` implementation produces the `props` to be passed to the `Feed` component which serves as the presentation layer, i.e. it produces the UI. Specifically, we're mapping the `fetchMore` prop to the following:
+
+```
+return fetchMore({
+  variables: {
+    offset: feed.length,
+  },
+  updateQuery: (prev, { fetchMoreResult }) => {
+    if (!fetchMoreResult.data) { return prev; }
+    return Object.assign({}, prev, {
+      feed: [...prev.feed, ...fetchMoreResult.data.feed],
+    });
+  },
+});
+```
+
+The `fetchMore` method takes a map of `variables` to be sent with the new query. Here, we're setting the offset to `feed.length` so that we fetch items that aren't already displayed on the feed. This variable map is merged with the one that's been specified for the query associated with the component. This means that other variables, e.g. the `limit` variable, will have the same value as they do within the component query.
+
+It can also take a `query` named argument, which can be a GraphQL document containing a query that will be fetched in order to fetch more information; we refer to this as the `fetchMore` query. By default, the `fetchMore` query is the query associated with the component, i.e. the `FEED_QUERY` in this case.
+
+When we call `fetchMore`, it will fire the `fetchMore` query and it needs to know how to incorporate the result of the query into the information the component is asking for. This is accomplished through `updateQuery`. The named argument `updateQuery` should be a function that takes the previous result of the query associated with your component (i.e. `FEED_QUERY` in this case) and the information returned by the `fetchMore` query and combine the two.
+
+Here, the `fetchMore` query is the same as the query associated with the component. Our `updateQuery` takes the new feed items returned and just appends them onto the feed items that we'd asked for previously. With this, the UI will update and the feed will contain the next page of items!
+
+Although `fetchMore` is often used for pagination, there are many other cases in which it is applicable. For example, suppose you have a list of items (say, a collaborative todo list) and you have a query that fetches only items that are added to this list. Then, you don't have to refetch the whole todo list: you can just incorporate the newly added items with `fetchMore` and specifiying the `updateQuery` function correctly.
 
 ## `updateQueries`
 
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -3,13 +3,13 @@ title: Updating your UI
 order: 7
 ---
 
-Within Apollo Client, we can specify a function through the `dataIdFromObject` option in order to give object identifiers to objects that our queries return. These allow Apollo Client to figure out when two different results actually point to exactly the same object. These object identifiers also allow Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. In particular, if you want to add something to a list of objects or if there are some objects that you can't assign an object identifier to, you should use either `fetchMore` or `updateQueries` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
+In most cases, your UI will be updated automatically as entries in Apollo Client's cache are updated. The function specified in the `dataIdFromObject` option can be used to assign a globally unique id to every object in the cache. These ids allow Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. For example, if you want to add something to a list of objects without refetching the entire list, or if there are some objects that you can't assign an object identifier to, Apollo Client cannot update existing queries for you. In those cases you have to use `fetchMore` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
 
 ## `fetchMore`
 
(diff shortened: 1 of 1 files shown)


source/cache-updates.md
```diff
@@ -3,9 +3,64 @@ title: Updating your UI
 order: 7
 ---
 
+Assigning object identifiers to your result objects using `dataIdFromObject` allows Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. In particular, if you want to add something to a list of objects or if there are some objects that you can't assign an object identifier to, you should use either `fetchMore` or `updateQueries` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
+
```
tmeasday: I think a sentence or two before this paragraph just reminding people about what the cache is and what it does might help a little bit.

tmeasday: Where are we documenting `dataIdFromObject` btw? Do we think it should be in here?

stubailo: I feel like we should have a special place where we document IDs/normalization/caching, and we should probably rename `dataIdFromObject` to something that makes more sense based on those docs? It's a really important concept.

source/cache-updates.md
```diff
@@ -3,9 +3,64 @@ title: Updating your UI
...
 ## `fetchMore`
 
-In order to incorporate the result of a query into the result of another query
+Typically, `fetchMore` is used to manually update the result of one query based on the data returned by another query. Most often, it is used to handle pagination. In our GitHunt example, we have a paginated feed that displays a list of GitHub respositories. When we hit the "Load More" button, we don't Apollo Client to throw away the repository information it has already loaded. Instead, it should just append the newly loaded repositories to the list of repositories that Apollo Client already has in the store. With this update, our UI component should re-render and show us all of the available repositories.
+
+This is possible with `fetchMore`. The `fetchMore` method allows us to fetch another query and incorporate that query's result into the result that our component query previously received. We can see it in action within the GitHunt code:
+
+```javascript
+const FeedWithData = graphql(FEED_QUERY, {
```
tmeasday: I think it'd be useful to show the query (or an abbreviated version)

source/cache-updates.md
```diff
@@ -3,9 +3,64 @@ title: Updating your UI
...
+  props({ data: { loading, feed, currentUser, fetchMore, variables } }) {
```
tmeasday: We don't need `variables` any more here I guess?

source/cache-updates.md
```diff
@@ -3,9 +3,64 @@ title: Updating your UI
...
+    return {
+      loading,
+      feed,
+      currentUser,
+      fetchMore() {
+        return fetchMore({
```
tmeasday: This function named `X` that calls a function named `X` thing is super confusing I bet. Can we think of a better way of structuring it?

source/cache-updates.md
```diff
@@ -3,9 +3,64 @@ title: Updating your UI
...
+  updateQuery: (prev, { fetchMoreResult }) => {
+    if (!fetchMoreResult.data) { return prev; }
+    return Object.assign({}, prev, {
+      feed: [...prev.feed, ...fetchMoreResult.data.feed],
+    });
+  },
+});
+```
+
+The `fetchMore` method takes a map of `variables` to be sent with the new query. Here, we're setting the offset to `feed.length` so that we fetch items that aren't already displayed on the feed. This variable map is merged with the one that's been specified for the query associated with the component. This means that other variables, e.g. the `limit` variable, will have the same value as they do within the component query.
```
tmeasday: We aren't showing values for variables in the component thought -- maybe we should?

source/cache-updates.md
tmeasday: Well this is an article about the cache so maybe this is the right spot?

source/cache-updates.md
Poincare: Yeah, I agree. This confused me while I was writing the example. Maybe we can change to `loadNextPage` as the outer call and it calls `fetchMore` internally?

source/cache-updates.md
stubailo: I am a bit concerned about using `feed.length` as the offset here - feels like we should be keeping track of the starting offset as well?

source/cache-updates.md
tmeasday: Yeah that makes sense to me. Something specific to the query, as opposed to the generic `fetchMore` works well.

source/cache-updates.md
Poincare: I don't think you can specify the starting offset in GitHunt anymore right (i.e. it is always 0)?

source/cache-updates.md
Poincare: Hm. Maybe it should be a separate section? I feel like this section should be about "here's what you should do if the automatic cache updating doesn't work for your particular usecase" which I think should be kept atleast somewhat separated from "here's the normal way: cache normalization."

source/cache-updates.md
tmeasday: Previously we used the old offset via `variables` (another named argument to `props()`), but this way seems cleaner to me?

stubailo: I guess this seems pretty limited only to cases where the starting offset is 0 and the total length is the next offset. I expect a lot of situations will require keeping track of temporary state, like a starting offset. but maybe we can add an example of that later

tmeasday: OTOH, this case is pretty typical, right? I think if we cover techniques to handle the more complex cases in the pagination section, this is good. cc @helfer

source/cache-updates.md
Poincare: fixed

source/cache-updates.md
Poincare: fixed

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
 order: 7
 ---
 
+Within Apollo Client, we can specify a function through the `dataIdFromObject` option in order to give object identifiers to objects that our queries return. These allow Apollo Client to figure out when two different results actually point to exactly the same object. These object identifiers also allow Apollo Client to reactively tell your queries about update your queries with results when new information becomes available. But, in some cases, just using `dataIdFromObject` is not enough for your application UI to get these updates. In particular, if you want to add something to a list of objects or if there are some objects that you can't assign an object identifier to, you should use either `fetchMore` or `updateQueries` in order to make sure that the queries on your page are updated with the right information and your UI updates correctly.
```
helfer: I'm not sure why this section starts with an explanation of `dataIdFromObject`. I think it would be better to just start by saying what `fetchMore` and `updateQueries` are for, and after that (if necessary) mention that in some cases you don't even need to use them.

helfer: Would be nice to add one line here that summarizes when to use `fetchMore` and when to use `updateQueries`, so people don't have to read both and then try to figure out which one to use.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+
 ## `fetchMore`
 
-In order to incorporate the result of a query into the result of another query
+Typically, `fetchMore` is used to manually update the result of one query based on the data returned by another query. Most often, it is used to handle pagination. In our GitHunt example, we have a paginated feed that displays a list of GitHub respositories. When we hit the "Load More" button, we don't Apollo Client to throw away the repository information it has already loaded. Instead, it should just append the newly loaded repositories to the list of repositories that Apollo Client already has in the store. With this update, our UI component should re-render and show us all of the available repositories.
```
helfer: I would mention that both of these queries, the "base query" and the "fetchMore query" are usually associated with the same component, otherwise it sounds like we're recommending bad coding practices.

"Most often it's used" could be "A common use case is ..."
GitHunt should be a link.

typo: "... we don't **want** Apollo Client ..."

Last sentence is superfluous, imho.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+
+This is possible with `fetchMore`. The `fetchMore` method allows us to fetch another query and incorporate that query's result into the result that our component query previously received. We can see it in action within the GitHunt code:
```
helfer: Make a link out of the GitHunt code that refers to the line (+ commit) where this example is from.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+
+```javascript
+const FEED_QUERY = gql`
```
helfer: How long is the feed query? If it's not super-long, it would be nice to see it all here.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+```
+
+We have two components here: `FeedWithData` and `Feed`. The `FeedWithData` implementation produces the `props` to be passed to the `Feed` component which serves as the presentation layer, i.e. it produces the UI. Specifically, we're mapping the `loadNextPage` prop to the following:
+
+```
+return fetchMore({
+  variables: {
+    offset: feed.length,
+  },
+  updateQuery: (prev, { fetchMoreResult }) => {
```
helfer: this is probably in GitHunt code, but I would call it `previousResult` instead of just `prev`. It's just clearer that way.

source/cache-updates.md
helfer: One problem I see here is that we're not telling people where `fetchMore` comes from, i.e. how to import it. Also, we say that fetchMore is a different query, but it seems that here it uses the same query, but just passes different variables. It might be useful to point out that fact.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+    return Object.assign({}, prev, {
+      feed: [...prev.feed, ...fetchMoreResult.data.feed],
+    });
+  },
+});
+```
+
+The `fetchMore` method takes a map of `variables` to be sent with the new query. Here, we're setting the offset to `feed.length` so that we fetch items that aren't already displayed on the feed. This variable map is merged with the one that's been specified for the query associated with the component. This means that other variables, e.g. the `limit` variable, will have the same value as they do within the component query.
+
+It can also take a `query` named argument, which can be a GraphQL document containing a query that will be fetched in order to fetch more information; we refer to this as the `fetchMore` query. By default, the `fetchMore` query is the query associated with the component, i.e. the `FEED_QUERY` in this case.
```
helfer: I see this is the explanation for where the query comes from. I think it would be useful to have that mentioned earlier. It would be easier to understand if the query was explicitly passed in `fetchMore`.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+
+When we call `fetchMore`, it will fire the `fetchMore` query and it needs to know how to incorporate the result of the query into the information the component is asking for. This is accomplished through `updateQuery`. The named argument `updateQuery` should be a function that takes the previous result of the query associated with your component (i.e. `FEED_QUERY` in this case) and the information returned by the `fetchMore` query and combine the two.
```
helfer: who is "it"? Apollo Client? You use "information" and "result" inconsistently here. I think it would be good to be very clear about what you mean exactly in each of those cases.

source/cache-updates.md
```diff
@@ -3,9 +3,68 @@ title: Updating your UI
...
+
+Here, the `fetchMore` query is the same as the query associated with the component. Our `updateQuery` takes the new feed items returned and just appends them onto the feed items that we'd asked for previously. With this, the UI will update and the feed will contain the next page of items!
+
+Although `fetchMore` is often used for pagination, there are many other cases in which it is applicable. For example, suppose you have a list of items (say, a collaborative todo list) and you have a query that fetches only items that are added to this list. Then, you don't have to refetch the whole todo list: you can just incorporate the newly added items with `fetchMore` and specifiying the `updateQuery` function correctly.
```
helfer: I think this paragraph could be improved by saying that you're only fetching items that were added after you made the initial query or something. Although the tricky part is how not to miss any data in between. Are we going to provide an example for that use case anywhere?

helfer:
I think it looks pretty good so far, but I'm not sure if it's really good if we use GitHunt as an example everywhere. If we do, we should have a section somewhere that explains GitHunt really well and that we can link to. If not, I think it would be easier for people to understand the concept of a blog with multiple authors and posts. GitHunt isn't something that people are familiar with. Even if you say it's like Hacker News, I think that's mostly a US thing, and our international developers might not understand it as readily.

One other thing: Where do we introduce react-apollo and how its functions work? I think that's probably the most important piece of this documentation if we use those functions in all the other examples. cc @tmeasday

Poincare:
I think we should just explain GitHunt carefully somewhere in that case. If we don't do this and just pick some other set of examples, we will be tasked with updating GitHunt all the time *and* updating the examples in our docs which probably means our docs will be constantly out of date.

source/cache-updates.md
Poincare: Agree w/ the import thing. Will fix.

I do try to make that distinction by describing the difference between the "fetchMore" query and the component query in the next paragraph. Is it not clear?

source/cache-updates.md
Poincare: @tmeasday suggested we should have a sentence to refresh the readers' mind on how the cache usually works. I do think it is very important to highlight the fact that these two functions are *not* the most common way of handling cache updates. 

Agree about the distinction between `updateQueries` and `fetchMore`. Will fix.

source/cache-updates.md
Poincare: I'm not sure if the "fetchMore query" is actually associated with a component at all. For example, you could have a component that displays all todo items and a fetchMore query that fetches only the newest todo items given a database cursor. If this is the case, the fetchMore query isn't really associated with the component but we still incorporate data returned by the fetchMore query into the component query.

source/cache-updates.md
tmeasday: We should probably fix the code as we make these changes too, if we feel it's easier to understand.

source/cache-updates.md
tmeasday: I think I was initially a little bit confused about that, but the text cleared it up for me. Maybe we should add a comment to the code or something to avoid the initial confusion.

stubailo:
The GitHunt schema is not terribly well designed, with types like `Entry`. I think for now it's probably fine but perhaps in the future we should rename some of that stuff to be a better example.

tmeasday:
> One other thing: Where do we introduce react-apollo and how its functions work? I think that's probably the most important piece of this documentation if we use those functions in all the other examples. cc @tmeasday

I'm thinking you read this before looking at the queries/mutations articles ;)

Personally I think there's a lot of value in seeing examples taken from as real an app as possible. That would be GitHunt in this case.

helfer:
Okay, I'll make some minor changes here and then merge this. As for GitHunt, I think we should work on making that a better/clearer example app, so I'll try to pick that up in the next focus cycle.

PR #25, opened 2016-08-23
SSR article

tmeasday:


source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
 order: 16
 ---
 
-XXX: just copied over from old docs, need to straighten out, and include http://docs.apollostack.com/apollo-client/index.html#store-rehydration
 
-The `react-apollo` library supports integrated server side rendering for both store rehydration purposes, or fully rendered markup. No changes are required to client queries to support this, however, queries can be ignored during server rendering by passing `ssr: false` in the query options. For example:
+Apollo provides two techniques to allow your applications to start quickly, avoiding unnecessary delays to users:
```
helfer: nitpick: I think "load quickly" would be better.

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+
+ - Store hydration, which allows your initial set of queries to execute immediately without a server roundtrip.
```
helfer: **re**hydration

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+ - Server side rendering, which allows you to serve the initial HTML the user sees directly.
```
helfer: ..., which renders the initial view on the server before sending it to the client.

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+
+```html
+<script>
+  // The contents of { ... } could be the result of client.store.getState(),
+  // or synthetically generated to look similar
+  window.__APOLLO_STATE__ = { ... };
+</script>
+```
+
+> We'll see below how you can generate both the HTML and the Apollo store's state using Node and `react-apollo`'s server rendering functions. However if you are rendering HTML via some other means, it might make sense to generate the state in some other way.
```
helfer: I think it would be better to say that you will most likely **have to** to generate the state in some other way if you're not already using react-apollo on the client, or Node on the server. However, the basic concept stays the same.

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+
+You can then rehydrate the client using the initial state passed from the server:
+```js
+const client = new ApolloClient({
+  initialState: window.__APOLLO_STATE__,
+});
+```
+
+> Note: if you are using [redux externally](redux.html) to Apollo, and already have store rehydration, you should pass the store state into the [`Store` constructor](http://redux.js.org/docs/basics/Store.html).
```
helfer: only "redux" should be linked, not "externally". Also, do they capitalize Redux or not?

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+
+Then, when a client calls runs queries, the data should be returned instantly because it is already in the store!
```
helfer: "should be" or "will be"?

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+
+> Note that if you are using [`forceFetch`](cache-updates.html#forceFetch) on queries, you should pass the `ssrForceFetchDelay` option to skip force fetching during initialization:
```
helfer: I think it's too early here to mention that, no? I don't think people will get the reason why. We should first show them how to do server side rendering or get the store state on the server, and then say: "oh, by the way, you'll have to pass ssrForceFetchDelay to make things work"

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
 ```js
 const withClientOnlyUser = graphql(GET_USER_WITH_ID, {
-  options: () => ({ ssr: false }) // won't be called during SSR
+  options: { ssr: false }, // won't be called during SSR
+});
+```
+
+<h3 id="server-initialization">Server Initialization</h3>
+
+When [creating an Apollo Client instance](initialization.html) on the server, you'll probably need to consider the network interface, and pass the `ssrMode: true` option to stop the server from force-fetching.
```
helfer: I think it would be valuable to explain in one or two sentences what apollo client does when you do ssr or store dehydration on the server. Then this sentence will make more sense to people. Right now it just sounds like an option people have to set without understanding why.

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+
+```js
```
helfer: What do you think about putting fake filenames and paths in all the code snippets so people have an idea of where that file goes and what it's for. It might also rub off some good naming conventions and repo organization practices.

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
+));
+
 ```
 
+> You should create a new client for each request, rather than re-using the same client for multiple requests as otherwise you'll have problems with [authentication](authentication.html) and you may see stale results.
 
 <h3 id="getDataFromTree">Using `getDataFromTree`</h3>
 
-The `getDataFromTree` function takes your React tree and returns the context of you React tree. This can be used to get the initialState via `context.store.getState()`
+The `getDataFromTree` function takes your React tree, determines and runs all queries required, and returns the context required to do so. This can be used to get the initialState via `context.store.getState()`
```
helfer: What do you mean "returns the context required to do so"? What's in the context? If it contains all the data that was fetched, that might be worth mentioning explicitly.

source/server-side-rendering.md
```diff
@@ -3,37 +3,154 @@ title: Server Side Rendering
...
 ```js
-// server application code (custom usage)
 import { getDataFromTree } from "react-apollo/server"
 
-// during request
+// during request (see above)
 getDataFromTree(app).then((context) => {
-  // markup with data from requests
+  // We are ready to render for real
   const markup = ReactDOM.renderToString(app);
```
helfer: should this be markup or content?

helfer:
I feel like I have a better idea of SSR and store re-hydration after reading this and could probably get it to work, but I think we could do a much better job explaining this - there are many places where new concepts are introduced abruptly, even before the previous concepts have been fully explained, especially at the headings. I kept expecting some more information to the concept that was just introduced, but instead a new concept was explained.

I'm not sure what the best way is to provide feedback on this stuff. I feel like it's kind of inefficient if I make a note for every tiny thing.

Going forward - if you agree - I will only make notes about major changes I think we should consider, and just edit the thing when I'd like to make a minor change in wording or sentence structure, or add some additional explanations. We can always revert changes if there's no agreement. I just think editing directly will reduce the back-and-forth for things that aren't super-important anyway.

source/server-side-rendering.md
tmeasday: Well you don't necessarily need to do SSR to use `initialState` (you could create the data with rails or something, maybe?)

tmeasday:
> I feel like I have a better idea of SSR and store re-hydration after reading this and could probably get it to work, but I think we could do a much better job explaining this - there are many places where new concepts are introduced abruptly, even before the previous concepts have been fully explained, especially at the headings. I kept expecting some more information to the concept that was just introduced, but instead a new concept was explained.

Do you think it just needs to be longer in general and explain SSR and how it all works more clearly? Or were there some specific things totally missing that it would be good to add.

> Going forward - if you agree - I will only make notes about major changes I think we should consider, and just edit the thing when I'd like to make a minor change in wording or sentence structure, or add some additional explanations. We can always revert changes if there's no agreement. I just think editing directly will reduce the back-and-forth for things that aren't super-important anyway.

Yes, please do!

source/server-side-rendering.md
tmeasday: It's a good idea but I'm not sure we have solid conventions yet? Are we happy w/ what we do in GH?

source/server-side-rendering.md
```diff
@@ -63,7 +61,17 @@ const withClientOnlyUser = graphql(GET_USER_WITH_ID, {
...
+
+We'll see how to take your component tree and turn it into a string in the next section, but you'll need to be a little careful in how you construct your Apollo Client instance on the server to ensure everything works there as well:
+
+1. When [creating an Apollo Client instance](initialization.html) on the server, you'll need to set up you network interface to connect to the API server correctly (this will look different to how you do it on the client).
+
+2. As you just want to fetch query results once, pass the `ssrMode: true` option to the Apollo Client constructor to avoid force-fetching.
+
+3. You need to ensure that you create a new client for each request, rather than re-using the same client for multiple requests as otherwise you'll have problems with [authentication](authentication.html) and you may see stale results.
+
+Once you put that all together, you'll end up with initialization code that looks something like (you can check out the [GitHunt app's `ui/server.js`](https://github.com/apollostack/GitHunt-React/blob/master/ui/server.js) for a complete working example):
```
helfer: I think the parenthesis got mixed up here?

source/server-side-rendering.md
```diff
@@ -3,37 +3,163 @@ title: Server Side Rendering
...
+const client = new ApolloClient({
+  initialState: window.__APOLLO_STATE__,
+});
+```
+
+> We'll see below how you can generate both the HTML and the Apollo store's state using Node and `react-apollo`'s server rendering functions. However if you are rendering HTML via some other means, you will have to generate the state manually.
+
+> Note: if you are using [Redux](redux.html) externally to Apollo, and already have store rehydration, you should pass the store state into the [`Store` constructor](http://redux.js.org/docs/basics/Store.html).
+
+Then, when a client calls runs queries, the data will be returned instantly because it is already in the store!
```
helfer: typo: calls or runs?

helfer:
This is awesome, really clear now!

PR #26, opened 2016-08-24
Basic redux docs

tmeasday:
I didn't include a lot of the thunk content from http://docs.apollostack.com/apollo-client/redux.html because it doesn't seem like you'd use it w/ `react-apollo`. Perhaps we want to keep it for the "raw" Client docs? I'm not sure.

### DIFF
--- a/source/redux.md
+++ b/source/redux.md
@@ -3,14 +3,12 @@ title: Integrating with Redux
 order: 15
 ---
 
-XXX: do we want to include James' stuff about MobX + React router?
 
-Explain how to use custom reducers for Apollo Client-related data
+By default, the Apollo Client creates its own internal Redux store to manage queries and their results. If you are already using Redux for the rest of your app, you can have the client integrate with your existing store instead. This will let you better track the different events that happen in your app, and how your client and server side data changes interleave.
 
-We should use http://docs.apollostack.com/apollo-client/redux.html
-and
+<h2 id="creating-a-store">Creating a Store</h2>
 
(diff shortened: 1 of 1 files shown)


stubailo:
Can we add this as well? https://github.com/apollostack/docs/pull/179

source/redux.md
```diff
@@ -3,14 +3,12 @@ title: Integrating with Redux
 order: 15
 ---
 
-XXX: do we want to include James' stuff about MobX + React router?
 
-Explain how to use custom reducers for Apollo Client-related data
+By default, the Apollo Client creates its own internal Redux store to manage queries and their results. If you are already using Redux for the rest of your app, you can have the client integrate with your existing store instead. This will let you better track the different events that happen in your app, and how your client and server side data changes interleave.
```
helfer: I feel like this is advice that you should have just one Redux store for your whole app, but it doesn't really go into why. Can we link to some resource that explains it?

helfer:
Looks fine to me so far. 👍

source/redux.md
stubailo: I think the main benefit is looking at your dev tools, and seeing one stream of events - also, if you're doing SSR/store hydration, you just have to send one blob to the client.

tmeasday:
> so far

Does that mean you think there should be more here? ;)

helfer:
Yep, I meant the dev tools stuff 😉

helfer:
Merged!

PR #27, opened 2016-08-24
Queries section

tmeasday:
The content came in a few commits, but this gives us one branch to comment on.

This reverts commit 73f18f64ffc5474d6ab7a3c122f86a5c829870e1.

### DIFF
--- a/source/queries.md
+++ b/source/queries.md
@@ -24,6 +24,7 @@ Profile.propTypes = {
   }).isRequired,
 };
 
+// We use the gql tag to parse our query string into a query document
 const CurrentUserForLayout = gql`
   query CurrentUserForLayout {
     currentUser {
--- a/source/queries.md
+++ b/source/queries.md
@@ -12,13 +12,19 @@ When we are using a basic query we can use the `graphql` container in a very sim
 For instance, in GitHunt, we want to display the current user (if logged in) in the `Profile` component:
 
 ```js
-import React, { Component } from 'react';
+import React, { Component, PropTypes } from 'react';
 import { graphql } from 'react-apollo';
 import gql from 'graphql-tag';
 
-function Profile({ data: { loading, currentUser } }) { ... }
+class Profile extends Component { ... }
+Profile.propTypes = {
+  data: PropTypes.shape({
+    loading: PropTypes.bool.isRequired,
+    currentUser: PropTypes.object,
+  }).isRequired,
+};
 
-const PROFILE_QUERY = gql`
+const CurrentUserForLayout = gql`
   query CurrentUserForLayout {
     currentUser {
       login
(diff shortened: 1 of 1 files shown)


source/queries.md
```diff
@@ -2,3 +2,182 @@
 title: Querying for data
 order: 3
 ---
+
+To fetch data from the server in a GraphQL system, we use GraphQL queries (you can read about the structure of GraphQL queries in detail at [graphql.org](XXX)).
+
+<h2 id="basics">Basic Queries</h2>
+
+When we are using a basic query we can use the `graphql` container in a very simple way. We simply need to parse our query into a GraphQL document using the [graphql-tag](../apollo-client/index.html#gql) library.
```
stubailo: I wouldn't focus on "parsing" to start with - I'd maybe just say "wrap your queries with the `gql` template tag" or something.

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+
+For instance, in GitHunt, we want to display the current user (if logged in) in the `Profile` component:
+
+```js
+import React, { Component } from 'react';
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+
+function Profile({ data: { loading, currentUser } }) { ... }
```
stubailo: It's not clear that this part of the code snippet represents a React component.

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+
+const PROFILE_QUERY = gql`
```
stubailo: Should the variable name be `CurrentUserForLayout`?

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+
+By default, `graphql` will attempt to pick up any missing variables from the query from `ownProps`. So in our example above, we could have used the simpler `withProfileData = graphql(PROFILE_QUERY);`. However, if you need to change the name of a variable, or compute the value, the `options` function is the place to do it.
+
+<h3 id="other-graphql-options">Other `options`</h3>
+
+Also, you may want to configure the [watchQuery](apollo-client-api.html#watchQuery) behaviour using `options`:
+
+```js
+const withPollingQuery = graphql(PROFILE_QUERY, {
+  options: () => ({ pollInterval: 1000 })
```
stubailo: Let's make sure to always use a big number for `pollInterval` in the examples - like 20 seconds. For anything smaller we should recommend subscriptions instead IMO.

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+<h2 id="graphql-props">Controlling child props</h2>
+
+As we've seen, by default, `graphql` will provide a `data` prop to the wrapped component with various information about the state of the query. We'll also see that [mutations](mutations.html) provide a callback on the `mutate` prop. However, if you want to decouple your presentational component from Apollo, and make it more reusable, you may want more control over the props passed into it.
+
+<h3 id="graphql-name">Using `name`</h3>
+
+If you want to change the name of this default property, you can use `name` field. In particular this is useful for nested `graphql` containers:
+
+```js
+import React, { Component } from 'react';
```
stubailo: No reason to import `Component` in these snippets.

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+    refetchUser: refetch,
+  }),
+});
+const ProfileWithData = withProfileData(Profile);
+```
+
+This style of usage leads to the greatest decoupling between your presentational component (`Profile`) and Apollo.
+
+
+<h2 id="default-result-props">The structure of the `data` prop</h2>
```
stubailo: Can we make it clearer how this list maps onto the arguments to the `props` option?

source/queries.md
helfer: So maybe add a comment that says that it's a component? Or extend `React.component` to make it super-clear.

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+  query CurrentUserForLayout {
+    currentUser {
+      login
+      avatar_url
+    }
+  }
+`;
+
+const withProfileData = graphql(PROFILE_QUERY);
+const ProfileWithData = withProfileData(Profile);
```
helfer: Why not just `graphql(PROFILE_QUERY)(Profile)`? I find this whole `withProfileData` vs. `ProfileWithData` business to be more confusing than useful. It's not like we use `withProfileData` anywhere else. And `withProfileData` really isn't the best name for what is essentially a function. Or is that a React convention that I'm not aware of?

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+```
+
+When we use `graphql` in this simple way with a GraphQL query document, the results are made available on a generic `data` prop on the child component (`Profile` in this case). In this case we can see that result object contains `loading`, a Boolean indicating if the the query is "in-flight", and (once the query has completed) `currentUser`, the field we've picked out in `PROFILE_QUERY`.
+
+We can expect the `data.currentUser` sub-prop to change as the logged-in-ness of the client and what it knows about the current user changes over time. That information is stored in Apollo Client's cache, and you can read more about techniques to bring the cache up to date with the server in the [article on the subject](cache-updates.html).
+
+<h3 id="graphql-options">Providing `options`</h3>
+
+If you want to configure the query, you can provide an `options` key on the second argument to `graphql`, and your options will be passed to [`ApolloClient.watchQuery`](apollo-client-api.html#watchQuery). In particular, if your query takes variables, this is the place to pass them in:
```
helfer: Why "in particular"? Could just drop that.

source/queries.md
```diff
@@ -2,3 +2,182 @@
...
+// And our HOC could look like:
+const withProfileData = graphql(PROFILE_QUERY, {
+  options: ({ avatarSize }) => ({ variables: { avatarSize } }),
+});
+const ProfileWithData = withProfileData(Profile);
+```
+
+By default, `graphql` will attempt to pick up any missing variables from the query from `ownProps`. So in our example above, we could have used the simpler `withProfileData = graphql(PROFILE_QUERY);`. However, if you need to change the name of a variable, or compute the value, the `options` function is the place to do it.
+
+<h3 id="other-graphql-options">Other `options`</h3>
```
helfer: So these options are presumably passed straight through to watchQuery. If that's the case, we should say that here.

source/queries.md
helfer: And maybe we could put this higher up? It seems to me like this would be one of the first things someone would want to know about the graphql function.

helfer:
I think this is pretty good. This will be one of the most important sections of the whole docs, so we should make sure it's perfect!

source/queries.md
tmeasday: Do you think? I was mildly confused about the `gql` incantation for a while and didn't know exactly what the point of it was (I still don't, TBH, why can't I pass in a string to things and you parse it for me? I'm lazy).

source/queries.md
tmeasday: Yeah, I was copying code out of GH (where we use the functional component style). I think `class Profile extends Component` is clearer, but the downside is we need to use the more verbose `propTypes` to explain the arguments it expects. Unless we think a simple render made-up function is better (than `propTypes`) to get this across?

source/queries.md
tmeasday: I could get behind that

source/queries.md
tmeasday: Hmm, yeah I think I agree. At some point we decided with @jbaxleyiii that it was a little bit simpler and less magic, but I'm not sure it's helping TBH.

source/queries.md
stubailo: > (I still don't, TBH, why can't I pass in a string to things and you parse it for me? I'm lazy).

Static analysis. We can pull out these strings at build-time if we need to later. Also, when we need to pass parsed queries into other stuff, we can use the independent `graphql-tag` package instead of loading up apollo client.

tmeasday: I'm not seeing how those benefits couldn't still apply if the `gql` call was on the other side of the API divide.

I guess to me using `gql` feels like an incantation that I blindly have to follow, kind of similar to the way Redux makes you call `combineReducers` rather than just accepting an array of reducers.

If there was some benefit to the user in having a parsed query (are there use cases where the user could pick out the variables or something?) it would make sense to me. I think the original comment you made here highlights why I don't really like this part of the API ("just remember to do this, don't worry about why").

tmeasday:
Ok, did another pass over this.

source/queries.md
stubailo: One more reason is that we can create a graphql-loader that pre-parses the queries at build time and passes them in. In general, I think having the GraphQL parsing/printing be separate from the core client implementation is good because you might not always need it, and it's worth having a special incantation that makes it trivial for any tool to come along and find all of your queries without having to guess at what is and isn't a GraphQL query. I'm not sure why it's bad to have a part of the API that exists purely for technical reasons - I haven't heard anyone complain that using GraphQL is annoying because you need to type a special tag at the beginning of query strings.

Oh, one more thing - we can use it to encourage people to avoid string interpolation and build any APIs we need that rely on the literal interpolation functionality (although I'm not a fan of this personally), because template literal tags are the only place where you get to work with the string segments before they are concatenated into one.

tmeasday: If there's a technical reason, then it's hardly onerous, which is probably why no one has complained (although as a user I was confused about it for a while, especially why I didn't have to use it when writing schemas in the schema language on the server). Just a (tiny) hurdle to get over that is worth examining.

I think the tooling example is a good reason to have it, although I'm still not sure we couldn't make our APIs just check the `query` argument to see if it's a string, and parse it if so. That would be the most flexible.

But if we are going to advocate using it (for the tooling reason), then it's pretty academic really, so I'll shut up now.

stubailo: It's one of those things that's easy to remove later, but _super_ hard to add. Which is why we decided to add it in a super breaking change in 0.3 or something. Let's keep it around for a while and if we don't find a good use for a few months we can take it out later.

helfer: I think it's easier to make the argument to add a `gql` requirement for the schema in graphql-tools than to remove it here. I think we should keep it until we have a good reason to remove it. It gives us one more layer, and I think the added flexibility is useful, especially this early in the project.

source/queries.md
```diff
@@ -2,3 +2,202 @@
...
+import gql from 'graphql-tag';
+
+class Profile extends Component { ... }
+Profile.propTypes = {
+  data: PropTypes.shape({
+    loading: PropTypes.bool.isRequired,
+    currentUser: PropTypes.object,
+  }).isRequired,
+};
+
```
helfer: Maybe add a comment here that explains the gql tag? Something like "We use the gql tag here to parse the query string.

tmeasday:
Ready to merge @helfer ?

helfer:
Yeah, looks good to me!

PR #30, opened 2016-08-29
Introduction

tmeasday:
Sashko I kind of wimped out on the comparison section, partly because I think you've thought about this a lot and should probably just write this.

Also:

 1. Should I expand the "goals" section? I think it mirrors what you've written before but with less detail

2. I removed the "Apollo is currently in the technical preview" paragraph. I think this fair?

3. The last index had a big list of examples. We will go into a lot of details about the GitHunt example in a section to come, I don't know if we want to mention the others still (or at least the React ones?)

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -3,6 +3,50 @@ title: Introduction
 order: 0
 ---
 
-## What it does
-## What it works with
-## How it compares to Relay / Lokka?
+This is the guide to using the [Apollo](http://apollostack.com) JavaScript client with the [React](https://facebook.github.io/react/) rendering library.
+
+Apollo is a collection of utilities designed to make it easier to use [GraphQL](http://graphql.org) across a range of front-end and server technologies. Although this guide focuses on the integration with React, there is a similar guide available for [Angular 2](XXX: link), and the `apollo-client` JavaScript package can be used in many other contexts too.
+
+If you are looking to use Apollo with a native mobile client, there is a [iOS Client](https://github.com/apollostack/apollo-ios) in development and plans for an Android client too. On the other hand, the React integration documented here works with [React Native](https://facebook.github.io/react-native/) on both platforms without changes.
+
+You can learn more about the Apollo project at the project's [home page](http://apollostack.com).
+
+<h2 id="apollo-client">Apollo Client and React</h2>
+
+The `apollo-client` npm module is a JavaScript client for GraphQL. The goal of the package is to be:
+
+1. **Incrementally adoptable**, so that you can drop it into any JavaScript app.
+2. **Simple to get started with**, you can just read one guide and get going.
+3. **Inspectable and understandable**, so that you can have great developer tools to understand exactly what is happening in your app.
+4. **Built for interactive apps**, so your users can make changes and see them reflected in the UI straight away.
+5. **Community driven**, many of the components of Apollo (including the `react-apollo` integration) were driven by our community and serve real-world use cases from the outset.
+
+The Apollo client does more than simply run your queries against your GraphQL server. It analyzes your queries and their results to construct a client-side cache of your data, which is updated as further queries, mutations are run and data is pushed to you from the server. This means that your UI can remain fully up-to-date with the state on the server with the minimum number of queries required.
+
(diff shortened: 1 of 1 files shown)


source/index.md
```diff
@@ -3,6 +3,50 @@ title: Introduction
 order: 0
 ---
 
-## What it does
-## What it works with
-## How it compares to Relay / Lokka?
+This is the guide to using the [Apollo](http://apollostack.com) JavaScript client with the [React](https://facebook.github.io/react/) rendering library.
```
stubailo: "JavaScript GraphQL client"? I guess people probably won't be confused but it can be a better summary I guess.

source/index.md
```diff
@@ -3,6 +3,50 @@ title: Introduction
...
+
+Apollo is a collection of utilities designed to make it easier to use [GraphQL](http://graphql.org) across a range of front-end and server technologies. Although this guide focuses on the integration with React, there is a similar guide available for [Angular 2](XXX: link), and the `apollo-client` JavaScript package can be used in many other contexts too.
```
stubailo: Rather than "Apollo is", perhaps we should say "The Apollo team builds and maintains..."

In some conversations about branding around optics, apollo server, etc, it makes the most sense to have "Apollo" refer to a set of people and a community, rather than a specific set of programs.

source/index.md
```diff
@@ -3,6 +3,50 @@ title: Introduction
...
+
+<h2 id="apollo-client">Apollo Client and React</h2>
+
+The `apollo-client` npm module is a JavaScript client for GraphQL. The goal of the package is to be:
+
+1. **Incrementally adoptable**, so that you can drop it into any JavaScript app.
+2. **Simple to get started with**, you can just read one guide and get going.
+3. **Inspectable and understandable**, so that you can have great developer tools to understand exactly what is happening in your app.
+4. **Built for interactive apps**, so your users can make changes and see them reflected in the UI straight away.
+5. **Community driven**, many of the components of Apollo (including the `react-apollo` integration) were driven by our community and serve real-world use cases from the outset.
```
stubailo: I'd add "planned and developed in the open" here - it's important that what we use internally is the same as what we give everyone else.

source/index.md
```diff
@@ -3,6 +3,50 @@ title: Introduction
...
+
+The best way to use `apollo-client` in your React app is with `react-apollo`, a React-specific API that's designed to take full advantage of Apollo. The integration provides a natural "higher-order-component" API for queries and mutations, and will keep your rendered component tree up to date with the data in the cache seamlessly.
+
+<h2 id="what-it-works-with">What it works with</h2>
+
+React Apollo provides a idiomatic React API, and is known to work in a straightforward way with many tools used in a typical React app. In particular:
+
+ - It works with React Native out of the box;
+ - It can integrate naturally with routers, such as React Router;
+ - Although it uses Redux internally (and you can [take advantage](redux.html) of this), you can easily use it with other state management tools like MobX.
```
stubailo: I'd actually say "it works really well with Redux, and _also_ with any other client-side data library that integrates with React". We don't need to apologize that we're Redux-optimized - I'd say that's a benefit for most people!

source/index.md
```diff
@@ -3,6 +3,50 @@ title: Introduction
...
+
+If you want to use Apollo with a different type of client (for instance a different JavaScript framework, or a native mobile app), it is possible too! Please check out the links at the top of this document, or head to our [developer site](http://apollostack.com) to get started.
```
stubailo: "different type of client" -> "different frontend environment"

stubailo:
> I removed the "Apollo is currently in the technical preview" paragraph. I think this fair?

Probably not that important to have it, since a fair number of people are using it in production. I think once we hit a more polished "1.0" release, we can make the wording even stronger in the "use this now it's going to be really great" direction.

I can handle the comparison section, just assign a task to me somewhere. But I think we can ship this without a detailed comparison.

As for examples, I think we need a page in the docs about "community examples". That way, it's clear which examples we update and support, but there's still an easy way to discover a ton of code you can browse for ideas.

------

Looks great, by the way!

tmeasday:
> As for examples, I think we need a page in the docs about "community examples". That way, it's clear which examples we update and support, but there's still an easy way to discover a ton of code you can browse for ideas.

I guess this is in the "pure" JS client docs, as opposed to the React client docs?

stubailo:
Could make sense to have a list specifically for React examples, but either one works for me. It could even be like `dev.apollodata.com/examples` on the website or something.

stubailo:
Pushed it 🔘

PR #33, opened 2016-08-31
Authentication / authorization pattern

### LINKED ISSUE #16: Authentication/Authorization

helfer:
Fixes #16


helfer:
Open questions:

* Where do we document all the other network interface stuff?
* Should we link to some docs that show how to deal with the authorization header on the server?
* Should I use the GitHunt example, or the current one (from the existing docs)

helfer:
@tmeasday circle is unhappy because of the code submodule not compiling. Did you run into the same problem earlier, or is this something new?

tmeasday:
Hmm, I think that's something new. I'll look at it.

helfer:
I was able to reproduce it locally after running `npm install` in the code submodule after running `submodule update`, so it may have to do with an updated version of some package. Although tsc is installed globally, right?

tmeasday:
@helfer, hmm no I think the submodule is trying to build itself so probably using it's own version of `tsc`. Let me know if you figure it out, otherwise I'll take a look in a minute

source/auth.md
```diff
@@ -1,4 +1,34 @@
...
+
+const networkInterface = createNetworkInterface('/graphql');
+
+networkInterface.use([{
+  applyMiddleware(req, next) {
+    if (!req.options.headers) {
+      req.options.headers = {};  // Create the header object if needed.
+    }
+    // get the authentication token from local storage if it exists
+    req.options.headers.authorization = localStorage.getItem('token') ? localStorage.getItem('token') : null;
```
tmeasday: This could be 

```js
req.options.headers.authorization = localStorage.getItem('token') || null;
```

tmeasday:
1. So I think the rest of the (more esoteric, I guess) network interface stuff should go in the "client core" docs I suppose.

2. This example isn't React specific at all, but I suppose it's important enough that it needs to be in the React guide. Is it worth extending the example to show both the server part (as you said) and maybe how the flow of your app when you login might look like?

3. I still feel like we should use githunt

4. Do we want to show examples of how to get the current user's information into your component hierarchy?

source/auth.md
```diff
@@ -1,4 +1,34 @@
 ---
-title: Authentication and Authorization
+title: Basic user authentication
 order: 10
 ---
+
+Some applications don't deal with sensitive data and have no need to authenticate users, but most applications have some sort of users, accounts and permissions systems. If different users have different permissions in your application, then you need a way to tell the server which user is associated with each request. Over HTTP, the most common way is to send along an authorization header.
+
+Apollo Client has a pluggable network interface that lets you modify requests before they are sent to the server.
```
tmeasday: We should link out to the rest of the docs about network interface here.

helfer: I don't think we have that part yet. Or do we?

helfer:
OK, I'll do 3. Not sure if 4 is necessary, because I don't see user information as being different from any other kind of information one might want.

helfer:
Oh, problem with 3 is that GitHunt just sends the cookie along, which it does with the `credentials` option to fetch. So maybe we should document that as well?

source/auth.md
tmeasday: No, but can we put a placeholder down?

tmeasday:
> I don't see user information as being different from any other kind of information one might want.

Sure but people might not think this way and prompting them that that's how they should think about it seems useful? Also how do you make sure it changes when they log in / out?

tmeasday:
> Oh, problem with 3 is that GitHunt just sends the cookie along, which it does with the credentials option to fetch. So maybe we should document that as well?

Sure!

helfer:
Made a couple of changes, I think I'll squash and merge now.

PR #38, opened 2016-09-02
Added a section about dataIdFromObject

### LINKED ISSUE #28: Explanation of `dataIdFromObject`and how the cache works
We need an explanation of how the cache works with object identifiers and query paths as well as an explanation of how to specify object identifiers through `dataIdFromObject`.

tmeasday:
Fixes #28

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -1,11 +1,44 @@
 ---
-title: Updating your UI
+title: Controlling the Cache
 order: 13
 ---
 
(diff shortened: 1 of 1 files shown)


source/cache-updates.md
```diff
@@ -1,11 +1,44 @@
 ---
-title: Updating your UI
+title: Controlling the Cache
```
stubailo: Maybe "updating the store"?

I think people will respond better to "store" than "cache".

tmeasday: Should we use the word "store" throughout, except maybe when defining what it is?

helfer: I'd be in favor of calling it cache everywhere, because that's really what it is. "Store" doesn't really mean anything other than "a place where you put some data".

stubailo: Cache sounds like a technical concept, store is what everyone in the react community calls it, especially when referring to flux.

helfer: Ok, so let's use "store" in the basic guide, but if we write more technical content, we should explain the cache aspect of the store.

PR #45, opened 2016-09-05
Redux connect mutation example update

rmarscher:
If I'm following the code properly, it needs to call `ownProps.onSelectList` instead of `ownProps.viewList`.

Thanks!


helfer:
Good catch. Thanks!

PR #46, opened 2016-09-05
gql import fix for mutations

rmarscher:
I'm pretty sure that `gql` needs to be imported as the default from graphql-tag instead of as a named import.


helfer:
Yes, you're right. Thanks!

PR #48, opened 2016-09-07
Update `props` example in queries.md

dangerismycat:
Update the code example syntax for using `props` in `graphql` to map props. The example as given errors.


helfer:
Thanks!

PR #57, opened 2016-09-14
Fix: the query options example is misleading

RyanCCollins:
Note: the example here was misleading:
```javascript
const FeedWithData = graphql(FeedEntries, {
  options: {
    return { pollInterval: 20000 };
  },
})(Feed);
```

I believe that it was supposed to be either a lambda that returns an object, or just a plain object without the `return` so I updated it to

```javascript
const FeedWithData = graphql(FeedEntries, {
  options: { pollInterval: 20000 },
})(Feed);
```


stubailo:
Good call, it was initially a lambda but I edited it incorrectly.

PR #71, opened 2016-10-07
Update server-side-rendering.md

acorcutt:
renderToStringWithData sends object { markup, initialState} in the callback not (content,state)
https://github.com/apollostack/react-apollo/blob/master/src/server.ts#L117


stubailo:
Good call, thanks!

PR #75, opened 2016-10-14
Added a note about how to do multiple mutations on a component.

GreenAsJade:
As requested (note: code fragment not tested, yet)

### DIFF
--- a/source/mutations.md
+++ b/source/mutations.md
@@ -58,6 +58,14 @@ const submitRepository = gql`
 const NewEntryWithData = graphql(submitRepository)(NewEntry);
 ```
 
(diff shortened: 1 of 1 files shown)


stubailo:
Two things:

1. Let's put the code snippet on two lines 
2. I think the default name is "mutate" not mutation

tmeasday:
Can we make the note also link to the part where the documentation for `graphql` is?

GreenAsJade:
@stubailo It's not entirely clear to me where I'd break that line :)

I'll suggest something :)

tmeasday:
Looks great. I'm not sure where is best to break the line either but this works.

PR #76, opened 2016-10-15
react-native.md: link to the "React guide"

lorensr:
With the idea that a common way people get to this page is clicking "React Native":

image

And they don't know they're already technically inside the React guide


stubailo:
Good call!

PR #77, opened 2016-10-18
Clarifications and more examples for "Controlling The Store" (dataIdFromObject and updateQueries)

GreenAsJade:
(courtesy of Tom@slack for how to)

Note: I would have liked to update the core-docs as well to fix up the apparent typo around addTypeName, and add more information about it there, then link to here.

But I couldn't quickly work out how to edit core-docs (%tsapibox%)

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -53,6 +53,8 @@ const apolloClient = new ApolloClient({
 }) 
 ```
 
+(Note: in 0.5.x, this will be the default behaviour)
+
 In some cases, just using `dataIdFromObject` is not enough for your application UI to update correctly. For example, if you want to add something to a list of objects without refetching the entire list, or if there are some objects that to which you can't assign an object identifier, Apollo Client cannot update existing queries for you.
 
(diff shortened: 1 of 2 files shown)


source/cache-updates.md
```diff
@@ -34,6 +34,25 @@ const client = new ApolloClient({
 
 These IDs allow Apollo Client to reactively tell all queries that fetched a particular object about updates to that part of the store.
 
+Apollo Client exports a method to make adding the `__typename` automatic: you provide this method as a transformer to the Apollo Client constructor.  So to do `dataIdFromObject` most concisely, your client initialisation might look like this:
```
stubailo: Funny story: in 0.5.x this is on by default! So I think we don't need these docs anymore :]

GreenAsJade: My feeling is you should merge this PR anyhow.

It's helpful for people who are still on 0.4.x.    It is consistent with the rest of the doc, which is till on 04.x.   

When the 0.5.x doc updates are happening, presumably this will be pulled out again, along with the comment in the preceding code example.

stubailo:
Good call. We'll update again for 0.5.

PR #78, opened 2016-10-18
fix error

avalla:
createNetworkInterface use an object as argument, read on [documentation ](http://dev.apollodata.com/core/apollo-client-api.html#createNetworkInterface)


stubailo:
@avalla we accidentally published a pre-release version with a breaking change, because we didn't realize `npm install` uses pre-releases by default. For now, please use `apollo-client@0.4.21` which is consistent with the docs.

Issue #80, opened 2016-10-18
Examples for multiple queries and mutators and custom scalar types


janikvonrotz:
Hey there,

Heard you would love to get some feedback.
I would have loved to see minimal example of multiple queries and mutators. Something like this:

```
...
const updateMutation = gql`
mutation updateCategory($id: String, $label: String) {
  updateCategory(id: $id, label: $label)
}
`;
const deleteMutation = gql`
mutation deleteCategory($id: String) {
  deleteCategory(id: $id)
}
`;
const CategoryEditWithMutation = graphql(updateMutation, {
  props({ mutate }) {
    return {
      updateCategory({id, label}) {
        return mutate({ variables: { id, label }});
      }
    };
  },
})(graphql(deleteMutation, {
  props({ mutate }) {
    return {
      deleteCategory(id){
        return mutate({ variables: { id }});
      }
    };
  },
})(CategoryEdit));
 
const query = gql`
query getCategory($id: String) {
  category(id: $id) {
    id
    label
  }
}
`;
const CategoryEditWithData = graphql(query, {
  options: (ownProps) => ({ variables: { id: ownProps.params.id } }),
})(CategoryEditWithMutation);
 
export default CategoryEditWithData;
...
```

another problem I've solved for my case was the definition of an custom scalar type. My example:

**schema.js**

```
scalar Date
...
type Category {
  id: String
  label: String
  createdAt: Date
}
...
```

Define a new scalar, which is basically another type.

**resolvers.js**

```
  Date: {
    __parseValue(value) {
      return new Date(value); // value from the client
    },
    __serialize(value) {
      return value.toISOString(); // value sent to the client
    },
    __parseLiteral(ast) {
      return ast.value;
    },
  },
```

Please let me know what you think of those ideas.

helfer:
@janikvonrotz thanks for the feedback! If it's not already documented, we should show how to combine multiple queries/mutations with `compose` (and maybe mention that one can also stack components like you did).

For the custom scalars, have you seen this? http://dev.apollodata.com/tools/graphql-tools/scalars.html

janikvonrotz:
@helfer Yes I could make a PR.
Is my example sufficient? 
Should I translate my example into your repository schema?
And in which chapter would like to see this kind of example?
=> It would be a mix of Usage / Queries and Usage / Mutations or should I seperate them?
=> I personally would have put it into the Mutations chapter.

thanks for your feedback.

helfer:
[...]

PR #81, opened 2016-10-18
multiple queries and mutations

janikvonrotz:
#80


helfer:
@tmeasday can you take a quick look at this? Do you think it's a reasonable recommendation, or should we tell people to use compose instead?

tmeasday:
I'm not really sure what I'd recommend. If you are using `@decorators` I think this approach works pretty well

```
@graphql(mutationOne)
@graphql(mutationTwo)
class Foo...
```

When you are using this style (non-decorators), all the nesting gets pretty confusing and it's easy to make mistakes. But `compose` just introduces another concept to understand, so I don't know if it is a net win.

source/mutations.md
```diff
@@ -194,3 +194,36 @@ When people talk about GraphQL, they often focus on the data fetching side of th
 In GraphQL, mutations can return any type, and that type can be queried just like a regular GraphQL query. So the question is - what type should a particular mutation return?
 
 In most cases, the data available from a mutation result should be the server developer's best guess of the data a client would need to understand what happened on the server. For example, a mutation that creates a new comment on a blog post might return the comment itself. A mutation that reorders an array might need to return the whole array.
+
+<h2 id="multiple-queries-and-mutation">Multiple queries and mutations</h2>
+
+As mentioned if you need more than one mutation on a component, you can make a container for each. Another solution would be to stack the containers together. As you know the compose function expects a component as parameter, instead of passing the raw component you can pass another container.
```
tmeasday: I don't really understand what this paragraph is trying to say. What would a container for each look like? Do you mean an intermediate named Component Class for each container?

Also, I'm not sure "As you know the compose function expects a component as parameter" is what you meant to say here?

source/mutations.md
```diff
@@ -194,3 +194,36 @@ When people talk about GraphQL, they often focus on the data fetching side of th
...
+        return mutate({ variables: { id, label }});
+      }
+    };
+  },
+})(graphql(query, {
+  options: (ownProps) => ({ variables: { id: ownProps.params.id } }),
+})(CategoryEdit));
+```
+
+Remember multiple queries can be lined up in one graphql tag. This would also work for mutations. However, in most cases one mutator call is expected to execute only one mutation.
```
tmeasday: I don't really understand this sentence either: "However, in most cases one mutator call is expected to execute only one mutation." -- what do you mean by this?

stubailo:
Please update according to @tmeasday's feedback if you have a chance @janikvonrotz!

helfer:
This is a bit too much out of sync with master now to merge, so I'm closing it.

PR #84, opened 2016-10-19
add webpack loader recipe

heldr:
As per https://github.com/apollostack/graphql-tag/pull/13

### DIFF
--- a/webpack.md
+++ b/webpack.md
@@ -1,0 +1,59 @@
+---
+title: Webpack loader
+---
+
+You can load GraphQL queries over `.graphql` files using Webpack. The package `apollo-tag` comes with a loader easy to setup and with many benefits:
+
+1. Do not process GraphQL ASTs on client-side
+2. Optimizes bundle size, since `apollo-tag` does not need to be imported
+3. Enable queries to be separated from logic
+
+In the example below, we create a new file called `currentUser.graphql`:
+
+```graphql
+query CurrentUserForLayout {
+  currentUser {
+    login
+    avatar_url
+  }
+}
+```
+
+You can load this file in 2 possible ways:
+
+<h2 id="using-connect">Default extension config</h2>
+
+You just need to add a rule in your webpack config file:
+
+```js
+loaders: [
+  {
+    test: /\.(graphql|gql)$/,
+    exclude: /node_modules/,
+    loader: 'graphql-tag/loader'
+  }
+]
+```
+
+As you can see, `.graphql` or `.gql` files will be parsed whenever imported:
+
+```js
+import React, { Component } from 'react';
+import { graphql } from 'react-apollo';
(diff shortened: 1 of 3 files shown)


webpack.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Webpack loader
+---
+
+You can load GraphQL queries over `.graphql` files using Webpack. The package `apollo-tag` comes with a loader easy to setup and with many benefits:
```
stubailo: The package is called `graphql-tag`.

webpack.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. Do not process GraphQL ASTs on client-side
+2. Optimizes bundle size, since `apollo-tag` does not need to be imported
```
stubailo: I think it's not immediately clear to me that for large apps the bundle size will be smaller. It depends on if the set of queries in the app is smaller or larger than `graphql-tag/parser`. But perhaps when combined with code splitting to only load the queries that are being used, it will always be better.

webpack.md
```diff
@@ -0,0 +1,59 @@
...
+
+class Profile extends Component { ... }
+Profile.propTypes = { ... };
+
+export default graphql(currentUserQuery)(Profile)
+```
+
+<h2 id="using-connect">Require hook</h2>
+
+This is less reusable because you need to prefix the loader for every query import:
```
stubailo: Is there any benefit to documenting this approach? It just doesn't seem like anyone should really be doing this.

webpack.md
heldr: Damn memory! Thanks! lol

webpack.md
heldr: Truth! Actually bundle size won't be smaller.

```graphql
query CurrentUserForLayout {
  currentUser {
    login
    avatar_url
  }
}
```
76 bytes

turns to:

```js
module.exports = {"kind":"Document","definitions":[{"kind":"OperationDefinition","operation":"query","name":{"kind":"Name","value":"CurrentUserForLayout"},"variableDefinitions":[],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","alias":null,"name":{"kind":"Name","value":"currentUser"},"arguments":[],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","alias":null,"name":{"kind":"Name","value":"login"},"arguments":[],"directives":[],"selectionSet":null},{"kind":"Field","alias":null,"name":{"kind":"Name","value":"avatar_url"},"arguments":[],"directives":[],"selectionSet":null}]}}]}}]}
```
658 bytes

stubailo: Actually if you just wanted the code separation you could make a loader that makes a module that parses on the client :)

let's take it one step at a time though.

PR #85, opened 2016-10-20
Clarify what version is documented.

GreenAsJade:


stubailo:
This seems kind of unnecessary - don't people have to intentionally install 0.5.0-1, since it's not used by npm automatically? It seems reasonable that docs are for the most recent stable version.

GreenAsJade:
I just did "npm install" in a clean directory and got:

├── apollo-client@0.4.22 extraneous
├── react@15.3.2 extraneous
└── react-apollo@0.5.13 extraneous


It's very confusing.

stubailo:
It's indicated by the version number, prerelease versions are a core feature of SemVer:

```
0.4.21 # stable
0.5.0-1 # unstable, has a SemVer pre-release
```

Perhaps we should call it something like: `0.5.0-preview.1` or something.

GreenAsJade:
I think it would be great to name it pre-release.     Would this also  persuade npm not to supply it as the default?   Or is it in fact that case that you want new people to get 0.5.x by default?

stubailo:
New people don't get 0.5 by default, as far as I know.

GreenAsJade:
As I mentioned above, I ran npm install react-apollo in a clean directory and got 0.5.13

stubailo:
Yeah it's extra confusing because there are multiple versions of many packages involved:

- `apollo-client`
- `react-apollo`
- `graphql-anywhere`
- `graphql-fragments`

And possibly more in the future.

GreenAsJade:
OK.  I definitely was seriously confused, and what I wrote in this PR is not only unnecessary, it is wrong.

Sorry about that.

GreenAsJade:
But ... what did someone mean when they said to me "in 0.5.x documentIdFromObject will be default behaviour".   I had thought they were referring to RA version 0.5.x, because that was the context: RA documentation.    I guess that comment needs to be clarified that it means AC 0.5.x!

PR #87, opened 2016-10-21
Add Relay-style cursor example to pagination

davidleach:
#52 

This PR adds an example of pagination using relay-style cursors and is based heavily on the example already provided for 'basic' cursor pagination.


stubailo:
Thanks, that's an awesome explanation!

PR #89, opened 2016-10-25
initial documentation for reducer option

helfer:


source/cache-updates.md
```diff
@@ -252,3 +252,49 @@ Note that the function must not alter the `prev` object (because `prev` is compa
 In our `updateQueries` function for the `Comment` query, we're doing something really simple: just adding the comment we just submitted to the list of comments that the query asks for. We're doing that using the `update` function from the `react-addons-update` package, just to do it concisely. But, if you wanted to, you could write some no-helper Javascript to combine the two incoming objects into a new one for the result.
 
 Once the mutation fires and the result arrives from the server (or, a result is provided through optimistic UI), our `updateQueries` function for the `Comment` query will be called and the `Comment` query will be updated accordingly. These changes in the result will be mapped to React props and our UI will update as well with the new information!
+
+
+<h2 id="resultReducers">Using `reducer`</h2>
+
+While `updateQueries` can only be used to update other queries based on the result of a mutation, the `reducer` option is a way that lets you update the query result based on any action, including results of other queries. It acts just like a Redux reducer on the non-normalized query result:
```
stubailo: I think perhaps we should recommend using `reducer` in most cases _instead_ of `updateQueries`?

source/cache-updates.md
```diff
@@ -252,3 +252,49 @@ Note that the function must not alter the `prev` object (because `prev` is compa
...
+    // ...
+  },
+  options({ params }) {
+    return {
+      reducer: (previousResult, action) => {
+        if (action.type === 'APOLLO_MUTATION_RESULT' && action.operationName === 'submitComment'){
+          return update(previousResult, {
+            entry: {
+              comments: {
+                $unshift: [newComment],
```
stubailo: Where did `newComment` come from?

source/cache-updates.md
helfer: I'd be okay with that, but then we should probably deprecate updateQueries.

PR #91, opened 2016-10-26
Change example code for skip option

JeroenBe:
# Change example code for `skip` option

Currently the docs state that `skip` is a field of the `options` object passed to `watchQuery`. Yet, the example code reflects otherwise. In the example code, `skip` is an object on the same level that `options` are. 

My PR changes the example code to reflect that `skip` is a part of the `options` object. 

I changed the docs as this is how it works right now (using `react-apollo^0.4.6`). Although, I find that `skip` should be a separate object, as it's not really an option to the query. In fact, the options may even be ignored entirely when `skip` is set to `true`, meaning it makes no sense to have `skip` be part of the object that's being ignored by the same skip.

[here's a link to the doc part I'm talking about](http://dev.apollodata.com/react/queries.html#graphql-skip)


stubailo:
In the newest version of React-apollo, it works both ways. 0.4 is not the latest version.

JeroenBe:
@stubailo Aha ok! My bad, is that something that should be documented? Or best just leave it as is?

helfer:
Let's leave it as it is. Our docs should always reflect the newest version (until we have separate docs for different versions post 1.0)

JeroenBe:
@helfer Yes that's totally fair, I was just ill informed :)

PR #92, opened 2016-10-28
passing as uri property

elitan:


helfer:
Both of these should work, btw. Did you test it?

stubailo:
@helfer we should accept this because the other way is deprecated.

elitan:
@stubailo Correct, I should have been more clear about that. Sorry.

PR #93, opened 2016-10-31
Document usage of compose

TimMikeladze:
Documented existence of compose function and provided usage example.


stubailo:
Great, thanks for adding this!

PR #95, opened 2016-10-31
subscribeToMore accepts document key for the query

guns2410:
Figured out looking at the [test](https://github.com/apollostack/apollo-client/blob/master/test/subscribeToMore.ts#L69) that subscribeToMore accepts the subscription query via the `document` key.


helfer:
Thanks, good catch!

PR #96, opened 2016-10-31
Update SSR doc to reflect change in createNetworkInterface

unkleho:
Got this notification following the docs:
Passing the URI as the first argument to createNetworkInterface is deprecated as of Apollo Client 0.5. Please pass it as the "uri" property of the network interface options.
Docs updated with new uri added and first argument removed.
ps: great work on Apollo!


stubailo:
Thanks! and thanks for the encouragement :]

PR #99, opened 2016-11-07
Add clarifications to compose documentations

TimMikeladze:
Relevant issue here: https://github.com/apollostack/react-apollo/issues/315

### DIFF
--- a/source/higher-order-components.md
+++ b/source/higher-order-components.md
@@ -113,11 +113,15 @@ const MyComponentWithUpvote = graphql(Upvote, {
 
 <h2 name='compose'>compose</h2>
 
-`react-apollo` exports a `compose` function. Adopting the following pattern allows you to reduce the number of reassignments you're doing every time you wrap your component with `graphql` and often `connect` from `react-redux`. 
+`react-apollo` exports a `compose` function. Adopting the following pattern allows you to reduce the number of reassignments you're doing every time you wrap your component with `graphql` and often `connect` from `react-redux`. Remember that the order in which you pass arguments to the `compose` function matters - the results of the rightmost function will be passed to the function on the left and so on. 
 
 ```js
 import { graphql, compose } from 'react-apollo';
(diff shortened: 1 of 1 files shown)


source/higher-order-components.md
```diff
@@ -113,17 +113,20 @@ const MyComponentWithUpvote = graphql(Upvote, {
 
 <h2 name='compose'>compose</h2>
 
-`react-apollo` exports a `compose` function. Adopting the following pattern allows you to reduce the number of reassignments you're doing every time you wrap your component with `graphql` and often `connect` from `react-redux`. 
+`react-apollo` exports a `compose` function. Adopting the following pattern allows you to reduce the number of reassignments you're doing every time you wrap your component with `graphql` and often `connect` from `react-redux`. Remember that the order in which you pass arguments to the `compose` function matters - the results of the rightmost function will be passed to the function on the left and so on.
```
stubailo: Can you clarify what this means for the container in particular? for example, if I want to get a variable out of Redux and pass it into `graphql`, what order should it be in? I see there's an idea in the example, but I think the reverse would be more interesting since I imagine most people will be passing local state as a variable into the mutation or query.

stubailo:
Ah, I didn't read the issue before. I think you might be encountering a bug, since IMO it should work in both directions.

source/higher-order-components.md
ucokfm: Yes i think we need second example on how to get a variable out of redux state and passing it to graphql by reversing arguments of compose function in first example. Because someone not familiar with redux compose would think that the rightmost function get result from the left function.

stubailo:
@TimMikeladze any updates on this PR? I posted a comment a while ago.

@ucokfm if you have time to pick this up and add the other example that would be awesome.

ucokfm:
yes, i have just created PR for that example.

helfer:
I think @calebmer 's recent PR clarified the use of compose very nicely.

PR #102, opened 2016-11-07
submit method return function

jakub-gawlas:
`submit` method should return function to invoke by onClick event

### DIFF
--- a/source/mutations.md
+++ b/source/mutations.md
@@ -116,7 +116,7 @@ import gql from 'graphql-tag';
 
 class NewEntry extends Component {
   render() {
-    return <div onClick={this.props.submit('apollostack/apollo-client')}>Click me</div>;
+    return <div onClick={() => this.props.submit('apollostack/apollo-client')}>Click me</div>;
   }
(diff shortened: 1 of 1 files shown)


stubailo:
Are you sure? I think it was correct before. The `props` function returns an object with a `submit` function. That function, when called, calls `mutate` with the right variables.

jakub-gawlas:
yes, but `submit` is used like that:
```
<div onClick={this.props.submit('apollostack/apollo-client')}>Click me</div>;
```
so `submit` is immediately invoked, another option is use `submit` like that:
```
<div onClick={() => this.props.submit('apollostack/apollo-client')}>Click me</div>;
```

stubailo:
Yeah, I don't think the first way is how people would usually use a callback. But we should add a code snippet of how `submit` would be called in the component (the second snippet) to the docs!

jakub-gawlas:
👍 changed

stubailo:
Looks great, thanks for catching this inconsistency!

PR #105, opened 2016-11-09
missing return in custom action case in reducer

kishorevarma:


stubailo:
Good catch, thanks!

kishorevarma:
Thanks

PR #113, opened 2016-11-20
Update docs to avoid getting two Redux stores

0xR:
If setStore is not called Apollo will create its own store as well as your redux store. This will result in a confusing experience in Redux devtools.

### DIFF
--- a/source/redux.md
+++ b/source/redux.md
@@ -36,8 +36,10 @@ const store = createStore(
   )
 );
 
+client.setStore(store);
+
 ReactDOM.render(
-  <ApolloProvider store={store} client={client}>
+  <ApolloProvider client={client}>
     <MyRootComponent />
   </ApolloProvider>,
   rootEl


helfer:
@jbaxleyiii do you know what the difference is here? I assumed that ApolloProvider would call `setStore`.

0xR:
Not sure. But if I didn't call setStore I ended up with 2 Redux instances
in the Redux devtools. Currently I can't investigate further, but it
shouldn't be too hard to reproduce.

Op di 13 dec. 2016 10:46 schreef Jonas Helfer <notifications@github.com>:

stubailo:
Hm, this shouldn't be the case - did you add the middleware to your existing redux store correctly?

helfer:
Thanks @0xR . After taking another look I think the example is correct. The store option on `ApolloProvider` just makes sure you also have access to the redux store in your components, not just the client instance. If this didn't work for you, please file an issue on react-apollo @0xR .

PR #114, opened 2016-11-20
Updated react fragment docs

0xR:
Using the gql helper in the fragment does not work because graphql-tag will compile the graphql to an object. This will result a query string like this:
```js
FeedEntry.fragments = {
  entry: gql`
    fragment FeedEntry on Entry {
      commentCount
      repository {
        full_name
        html_url
        owner {
          avatar_url
        }
      }
      ...VoteButtons
      ...RepoInfo
    }
    [object Object]
    [object Object]
  `,
};
```

Since this is not valid graphql you get an syntax error.

### DIFF
--- a/source/fragments.md
+++ b/source/fragments.md
@@ -35,7 +35,7 @@ To do so, we can simply share a fragment describing the fields we need for a com
 import gql from 'graphql-tag';
 
 CommentsPage.fragments = {
-  comment: gql`
+  comment: `
     fragment CommentsPageComment on Comment {
       id
       postedBy {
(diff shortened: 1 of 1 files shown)


stubailo:
@0xR are you updated to the newest version of graphql-tag?

0xR:
Updating graphql-tag fixes it. I was on 0.1.16 and a couple of days ago 0.1.17 was released and today 1.0.0 was released.

Was the fix in 0.1.17? or was it in 1.0.0 which was not released when I submitted the issue. That would mean the docs were actually ahead of the graphql-tag release which is confusing.

PR #115, opened 2016-11-20
Add missing closing backticks to fragment example

SimenB:
To avoid 
image


stubailo:
Good call, I just fixed this in another commit.

PR #116, opened 2016-11-21
queryTransformer no longer supported

kishorevarma:
it would be misguiding new users

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -41,16 +41,16 @@ const client = new ApolloClient({
 
 These IDs allow Apollo Client to reactively tell all queries that fetched a particular object about updates to that part of the store.
 
-Apollo Client exports a method to make adding the `__typename` automatic: you provide this method as a transformer to the Apollo Client constructor.  So to do `dataIdFromObject` most concisely, your client initialization might look like this:
+You can make `__typename` automatic by adding addTypename property as true to the Apollo Client constructor. So to do `dataIdFromObject` most concisely, your client initialization might look like this:
 
 ```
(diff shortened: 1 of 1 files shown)


source/cache-updates.md
```diff
@@ -41,16 +41,16 @@ const client = new ApolloClient({
...
 ```
-import ApolloClient, {createNetworkInterface, addTypename} from 'apollo-client'
+import ApolloClient, {createNetworkInterface} from 'apollo-client'
 
 const networkInterface = createNetworkInterface('http://localhost:3000/graphql') // TBD: Need to provide the right path for production
 
 const apolloClient = new ApolloClient({
     networkInterface: networkInterface,
-    queryTransformer: addTypename,
+    addTypename: true,
```
stubailo: This is on by default - so perhaps we can simply remove this section, or explain why someone might want to set `addTypename: false`.

kishorevarma: yeah that would be good,  and its explained why we need to use , so I don't think we need to explain why we should not use, its my opinion 

> // If the IDs are only unique per type (this is typical if an ID is just an
// ID out of a database table), you can use the `__typename` field to scope it.
// This is a GraphQL field that's automatically available, but you do need
// to query for it also

stubailo:
If you have time please update the PR!

source/cache-updates.md
kishorevarma: I will modify it, and remove the section.

helfer:
Sorry, i didn't see this PR and just removed the offending lines today.

PR #117, opened 2016-11-21
Fix missing closing code block

shanestillwell:


shanestillwell:
guess there are several others that PR'd the same thing.

PR #121, opened 2016-11-28
Update server-side-rendering.md

judikdavid:
Fix createNetworkInterface initialization, add options as an object in the second parameter


stubailo:
Thanks!

PR #122, opened 2016-11-30
Remove unneeded backtick in cache-updates.md

kitze:
Probably just a typo. 
Also, I'm not sure where the ```dataIdFromObject``` should be imported from.


stubailo:
Good call - @helfer, any thoughts about how best to do this? Either way we should make the example more complete.

@kitze I'd say you can just import the one you provide to Apollo Client under the `dataIdFromObject` option.

PR #124, opened 2016-12-01
Fix a few code examples

fson:
* Mutations require a selection set.
* Change `’true’` string to `true` boolean.
* Fix the props of `List` in the Redux example.

### DIFF
--- a/source/redux.md
+++ b/source/redux.md
@@ -70,7 +70,9 @@ import { connect } from 'react-redux';
 import { CLONE_LIST } from './mutations';
 import { viewList } from './actions';
 
-const List = function({ listId, onSelectList });
+const List = ({ listId, cloneList }) => (
+  <div>List ID: {listId} <button onClick={cloneList}>Clone</button></div>
+);
 
 const withCloneList = graphql(CLONE_LIST, {
   props: ({ ownProps, mutate }) => ({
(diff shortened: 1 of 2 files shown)


source/redux.md
```diff
@@ -70,7 +70,9 @@ import { connect } from 'react-redux';
 import { CLONE_LIST } from './mutations';
 import { viewList } from './actions';
 
-const List = function({ listId, onSelectList });
+const List = ({ listId, cloneList }) => (
```
fson: I looked at the code below and `onSelectList` doesn't make sense as prop because `cloneList` is defined in the `props` function passed to the container. I searched the git history and it turns out it had `cloneList` in the beginning, but was [changed](https://github.com/apollostack/react-docs/commit/f61f9a0fd32900f26e0ac54770c218aa2cebc89a#diff-db2186b46f16eab630429b8dcb268828R70) by mistake at some point.

` function({ listId, onSelectList });` is also not valid syntax, so I expanded the example a bit.

stubailo:
> Mutations require a selection set.

That's only the case if they return an object type. Relay mutations always do but it's not a requirement in GraphQL in general.

I do think most mutations will return object types though so let's put that in there!

stubailo:
Thanks for the improvements.

Issue #59, opened 2016-09-17
Make sure it's clear how to get from mutations to updating the store


stubailo:

chaffeqa:
Some items pertaining to this issue I have recently encountered (fairly new to typescript so its a bit hard for me to debug, sorry! )

## Lifecycle:
This is important for doing things with forms, where you want to *only* copy state from the data after it has been updated (via refetch or update) after a submit.

- When does `mutate.then` in context to the following:
- refetchQueries
- updateQueries


## updateQueries
Ran into an issue (bug? documentation?) where if you have 2 queries to update based on a mutation:

```
gql`
mutation setDailyLineup {
  setDailyLineup{
    dailyLineups {
      id
      ...
    }
    dailyContest {
      id
      ...
    }
  }
}
`
```

if you *dont* specify an `updateQueries`, the non-list `dailyContest` successfully updates, however if we *do* implement an `updateQueries`:

```
        updateQueries: {
          dailyLineups: (prev, { mutationResult }) => {
            if (mutationResult.data.setDailyLineup) { // result has errors?
              const addedLineup = mutationResult.data.setDailyLineup.dailyLineup
              if (!prev.dailyLineups.find((lineup) => lineup.id == addedLineup.id)) {
                console.log(`adding to dailyLineups: ${addedLineup.id}`);
                return update(prev.dailyLineups, { $unshift: [addedLineup] });
              }
            }
            return prev
          },
        },
```

The `dailyContest` never gets updated.

My hunch is that when you override the default `updateQueries` you are responsible for *all* the queries

marktani:
Not sure if I should create a new issue, but I think this fits here.

I got updating the store to work after creating a new node by using the following approach, and I am not fully understanding why this **does** work:

* I have a query to fetch all books of an author
  ```
  Author("Tolkien") {
    books {
      title
    }
  }
  ```

* When creating a new book, including the author id and the books id in the mutation response makes Apollo update the list of books:

  ```
  createBook(title: "Title" author: "tolkien-id") {
    author {
      id
      books {
        id
      }
    }
  }
  ```

  If I leave any of the two `id` fields out, the store will not automatically update. To make this work, of course I also have to use `dataIdFromObject`.

Relevant part of the `updateQueries` chapter:

> However, if you are removing or adding items to a list with a mutation or can’t assign object identifiers to some of your objects, you’ll have to use updateQueries to make sure that your UI reflects the change correctly.

helfer:
[...]

PR #126, opened 2016-12-02
Added mention to mutation triggered query refetch/update.

lucasconstantino:
Both mutation options are related to the subject of getting updates from the server.


tmeasday:
Maybe we can link to http://dev.apollodata.com/react/cache-updates.html#updateQueries though?

lucasconstantino:
Sure! No detailed documentation on the refetchQueries though, right?

This is related to #59 and #58

tmeasday:
Thanks!

PR #147, opened 2017-01-28
Remove reference to future subscriptions feature

jessedijkstra:
Remove reference to future subscriptions feature which is already available in the Apollo client.


helfer:
Thanks @jessedijkstra !

PR #148, opened 2017-01-31
change implied query name to reduce confusion

srtucker22:
I think it would be helpful to make it clear that params of `updateQueries` first arg are the names of queries. 

In the example in the docs, there is an implied query named Comment, but the type Comment is also returned by the `submitComment` mutation. It wasn't clear to me that the param refers to queries and not the return type of the mutation, especially since the `optimisticResponse` params are all about type. Took investigating the docs to realize why updateQueries wasn't executing.

The easiest way I can think of showing this without making the example more complex is just renaming the query :)

### DIFF
--- a/source/optimistic-ui.md
+++ b/source/optimistic-ui.md
@@ -91,7 +91,7 @@ const CommentsPageWithMutations = graphql(SUBMIT_COMMENT_MUTATION, {
             },
           },
           updateQueries: {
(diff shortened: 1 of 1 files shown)


stubailo:
Why don't we just name it `CommentQuery`? And perhaps add something like:

```js
// Would update the query that looks like:
// query CommentQuery { ... }
```

srtucker22:
That's way clearer. 🍺

PR #149, opened 2017-02-07
updated webpack.md with info on #import syntax for fragments

jole78:
see discussion in https://github.com/apollographql/graphql-tag/pull/33


stubailo:
Thanks a lot!

PR #151, opened 2017-02-10
Nest return value behind data key

jonmagic:
When I tried to make this work in a project I got the following error:

Warning: Failed prop type: The prop `data` is marked as required
Uncaught TypeError: Cannot read property 'loading' of undefined

Screenshot: https://cl.ly/1E0i301t452N

Maybe my component is wired up wrong though? I fixed the problem by returning a hash with the data key and everything else nested inside of it.


stubailo:
I think the intention is that you would use those props in your component directly rather than having one big data prop.

jonmagic:
👍🏼

stubailo:
One way to improve the docs here would be to add an example of a React component to consume that data, then it would be clear that those fields match up to individual props!

PR #153, opened 2017-02-12
Update Redux devtools setup syntax

WebDevFromScratch:
As https://github.com/zalmoxisus/redux-devtools-extension#usage docs states, `window.devToolsExtension` has been renamed and will be deprecated in the future. I just stumbled upon this, while setting up my latest project, so I'm not sure if I can just propose a change like this here.. Let me know in case I'm trying to contribute in a wrong way.

And keep up the great work, Apollo Client is awesome!


helfer:
Thanks for the PR @WebDevFromScratch ! 🙂

PR #156, opened 2017-02-14
add a note regarding reducer for custom redux actions

callmephilip:
it looks like reducer feature has been broken for some time now. adding a note here - https://github.com/apollographql/apollo-client/issues/1013


helfer:
Thanks @callmephilip ! I made a few more changes, but I think it's important to add the note there.

PR #157, opened 2017-02-15
Update initialization.md

nikolasburk:
fixed typo


stubailo:
How is this a typo? Doesn't `ReactDOM.render` require an element to render into?

helfer:
I think it looked like someone forgot to finish writing "rootElement". How about we just call it `domContainerNode`, which is what the official React website puts there. 🙂

PR #160, opened 2017-02-20
use typeof to check for redux devtools

zebapy:
The original `window.__REDUX_DEVTOOLS_EXTENSION__ && ...` was breaking my app in browsers without this extension.

I'm not sure how other parts of my app are influencing the problem that is caused by this but I can't be alone. I'm using create-react-app not ejected.


stubailo:
Thanks!

PR #161, opened 2017-02-21
Docs: add link to the ssr branch on GitHunt-React

RyanCCollins:
When I was integrating ssr into my project, I found that I wanted a more complete example and I ended up on the GitHunt-React repository.  I figured linking to it could help others, as well.

Also, I was curious as to what the difference between the getDataFromTree function and the [renderToStringWithData](https://github.com/apollographql/GitHunt-React/blob/master/ui/server.js#L59)?  I don't know enough about this to add the latter to the docs.  Should this also be added?


RyanCCollins:
Edit: I think that the docs might be out of date here.  Should it be updated to the example shown in the [master branch](https://github.com/apollographql/GitHunt-React/blob/master/ui/server.js#L59) of the Githunt-React repo, rather than what's in the [ssr branch](https://github.com/apollographql/GitHunt-React/blob/ssr/ui/server.js#L54)?  I'd be happy to update it, but am not sure on what's changed.

RyanCCollins:
Closing since I realize now that the ssr example on githunt is out of date.  I will get to the bottom of this and update the docs when I have a better understanding of the internals of React-Apollo.

PR #163, opened 2017-02-24
Add documentation for https://github.com/apollographql/react-apollo/pull/481

1st8:
https://github.com/apollographql/react-apollo/pull/481


helfer:
Since the other PR is closed, I'll close this as well. Let's keep the conversation going though @1st8, because you're definitely not the only one with this use case. We should just make sure to find a nice API that enables it. 🙂

PR #165, opened 2017-03-03
Remove outdated note about warning

tmeasday:
See https://github.com/apollographql/apollo-client/issues/940


helfer:
Thanks @tmeasday ! 👍

PR #166, opened 2017-03-03
API documentation

calebmer:
This adds API documentation for `react-apollo` so that users have a reference of everything they can do with the bindings. Unlike the API documentation for `apollo-client` this does not use `typedoc` to generate documentation. I think we can do a better job communicating everything we need to communicate through a hand-written doc. We can have more code examples, and better document things specific to React like props.


stubailo:
This looks like the right direction for me! I agree that TypeDoc makes it harder to add human flavor.

source/api.md
```diff
@@ -0,0 +1,194 @@
...
+  query TodoAppQuery {
+    todos {
+      id
+      text
+    }
+  }
+`)(TodoApp);
+```
+
+You may also be interested in using the `graphql` function as a [decorator][] on your React class component.
```
helfer: I would say "Alternatively, you can also use the `graphql` function ..."

source/api.md
```diff
@@ -0,0 +1,194 @@
+---
+title: API Documentation
+---
+
+<h2 id="graphql">`graphql(query, config)(component)`</h2>
+
+The `graphql` function is the most important thing exported by `react-apollo`. With this function you can create higher-order components that can execute queries and update reactively based on the data in your Apollo store. The `graphql` function creates an “enhancer” function which when called with a component the enhancer function will create a new component with reactive GraphQL capabilities. This follows the React [higher-order component][] pattern which is also used by [`react-redux`’s `connect`][] function.
```
helfer: The "enhancer" sentence of this paragraph doesn't read that well, especially around "... which when called with ..."

source/api.md
```diff
@@ -0,0 +1,194 @@
...
+        ))}
+      </ul>
+    );
+  }
+}
+```
+
+In this guide, we won't use the decorator syntax to make the code more familiar, but you can always use it if you prefer.
+
+The configuration options that your `graphql` function accepts will be different depending on if your GraphQL operation is a [query](#queries) or a [mutation](#mutations). Go to the appropriate API documentation for more information about available options.
```
helfer: I believe it also takes subscriptions?

source/api.md
```diff
@@ -0,0 +1,194 @@
...
+        {todos.map(({ id, text }) => (
+          <li key={id}>{text}</li>
+        ))}
+      </ul>
+    );
+  }
+}
+```
+
+In this guide, we won't use the decorator syntax to make the code more familiar, but you can always use it if you prefer.
```
helfer: I think this sentence isn't necessary.

source/api.md
```diff
@@ -0,0 +1,194 @@
...
+```js
+render() {
+  const { data } = this.props;
+
+  console.log(data.viewer); // <- The data returned by your query for `viewer`.
+  console.log(data.todos); // <- The data returned by your query for `todos`.
+}
+```
+
+The `data` prop has some other useful properties which are all documented below. These properties can be accessed directly from `data`. For example, `data.loading` or `data.error`. These properties are documented below.
```
helfer: "documented below" appears twice.

helfer: Maybe we should tell people to always check `data.loading` and `data.error` in their components, or code in such a way that they don't assume `data` is there?

source/api.md
```diff
@@ -0,0 +1,194 @@
...
+
+<h3 id="graphql-query-data.error">`props.data.error`</h3>
+
+If an error occurred then this property will be an instance of [`ApolloError`][]. If you do not handle this error you will get a warning in your console that says something like: `"Unhandled (in react-apollo) Error: ..."`.
+
+[`ApolloError`]: /core/apollo-client-api.html#ApolloError
+
+<h3 id="graphql-query-data.variables">`props.data.variables`</h3>
+
+The variables that Apollo used to fetch data from your GraphQL endpoint. This property is helpful if you want to render some information based off of the variables that were used to make a request against your server.
```
helfer: "based on"

source/api.md
```diff
@@ -0,0 +1,194 @@
...
+- `[variables]`: The optional variables you may provide that will be used with either the `query` option or the query from your `graphql` HOC (depending on whether or not you specified a `query`).
+- `updateQuery(previousResult, { fetchMoreResult, queryVariables })`: This is the required function you define that will actually update your paginated list. The first argument, `previousResult`, will be the previous data returned by the query you defined in your `graphql` function. The second argument is an object with two properties, `fetchMoreResult` and `queryVariables`. `fetchMoreResult` is the data returned by the new fetch that used the `query` and `variables` options from `data.fetchMore`. `queryVariables` are the variables that were used when fetching more data. Using these arguments you should return a new data object with the same shape as the GraphQL query you defined in your `graphql` function. See an example of this below, and also make sure to read the [pagination](pagination.html) recipe which has a full example.
+
```
helfer: What about `refetch` and `subscribeToMore`? We should make sure we document the entire API of the `graphql` function here, because there's no other place where it's documented in its entirety.

source/api.md
calebmer: Copied it over from the old documentation 😉

source/api.md
calebmer: Yep. I don’t know enough about subscriptions to write the documentation though, so I excluded it. I also thought we were going to do some work on the subscriptions API. I’ll defer this section to @Urigo.

source/api.md
calebmer: I had to go eat lunch. I didn’t finish 😊

calebmer:
Incorporated your suggestions @helfer. I’ll ping this thread again when it’s done or there is enough progress to start looking at it again 👍

calebmer:
**Update:** There is a lot to document 😊

So far I’ve documented all of the general config options, query data methods, and query options. I’m now moving on to document mutations. After that we’ve got all the other things exported by `react-apollo` like `withApollo`, `compose`, `ApolloProvider`, `getDataFromTree`, and `renderToStringWithData`.

I haven’t been documenting the subscription stuff because last time I heard we were planning on changing the subscriptions API in `react-apollo` for 1.0. I’m also not familiar with the subscriptions API. @Urigo, do you want to fill in the subscription API docs?

After the API reference is done I’m going to revise some of the other documentation articles to reference the API docs and cleanup any partial API docs that may exist in those articles.

stubailo:
This is awesome, thanks for the update!

calebmer:
Just a note, there is a bug in the sidebar documentation. Because we use a heading of `<ApolloProvider client={client} />` for the `ApolloProvider` the sidebar doesn’t properly escape the `<` and `>` characters. What would be the appropriate repo to PR to fix this bug?

calebmer:
Also, I’m not documenting a couple of things which may be worth discussing:

- https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L48: Not sure what it does.
- https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L52: Do we even want people to use this?
- https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L55: It’s being removed.
- https://github.com/apollographql/react-apollo/blob/master/src/graphql.tsx#L59: Deprecated.

stubailo:
- `forceFetch` on mutation... that's certainly curious. I don't think it does anything, I bet the definition is just out of date.
- `ssr` option - people asked for a feature to ignore queries during SSR, but I can't track down the discussion. @tmeasday or @jbaxleyiii might know more.
- `returnPartialData`, `skip` - great. No need to document.

calebmer:
@stubailo, @helfer all done with the full API reference. If you can start reading and reviewing that document it would be awesome 👍

If you’ve started reading before, I suggest you re-read because things were added and moved around.

I also added brief documentation with external links for `gql`, `ApolloClient`, and `createNetworkInterface` as they are not exported by `react-apollo`. Ideally if you are a `react-apollo` user that document should be the only documentation you need open while you develop and you can follow links from there when the documentation is insufficient.

To view the rendered document clone the repo and run `npm install` and then `npm start` and navigate to `/react/api.html`.

I still haven’t documented subscriptions because I’m not sure if they are going to change or not, and I’m not the most qualified person to write those docs. Let me know how you want to proceed on subscriptions.

I’m now going to make some changes to the other documentation articles to get them to reference this reference and reflect other changes in the API since they were initially written. In the meantime definitely start reviewing the reference document 👍

source/api.md
```diff
@@ -939,10 +941,101 @@ ReactDOM.render(
...
+function MyComponent({ client }) {
+  console.log(client);
+}
+```
+
+<h2 id="compose">`compose(...enhancers)(component)`</h2>
+
+For utility purposes, `react-apollo` exports a `compose` function. Using this function you may cleanly use several component enhancers at once. Including multiple [`graphql()`](#graphql), [`withApollo()`](#withApollo), or [Redux `connect()`][] enhancers. This should clean up your code when you use multiple enhancers. [Redux][] also exports a `compose` function, and so does [Recompose][] so you may choose to use the function from whichever library feels most appropriate.
+
+An important note is that `compose()` non-intuitively executes the last enhancer _first_ and works its way backwards through the list of enhancers. If this does not make sense to you consider using [`flowRight()` from Lodash][] which otherwise has the same behavior.
```
stubailo: I actually don't think this is as unintuitive as it sounds, right?

You can think of:

```js
funcA(funcB(funcC(component)))
```

Mapping onto:

```js
compose(funcA, funcB, funcC)(component)
```

So if you think about the execution order I agree it's not intuitive, but if you think about the code it's replacing, then it seems like the most straightforward thing.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+```js
+const fragments = gql`
+  fragment foo on Foo {
+    a
+    b
+    c
+    ...bar
+  }
+
+  fragment bar on Bar {
```
stubailo: I believe this code sample will actually cause an error due to an unused fragment? Or am I missing something?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+        ))}
+      </ul>
+    );
+  }
+}
+```
+
+The `graphql()` function will only be able to provide access to your GraphQL data if there is a [`<ApolloProvider/>`](#ApolloProvider) component higher up in your tree to actually provide the client and network interface that will be used to fetch your data.
+
+The behavior of your component enhanced with the `graphql()` function will be different depending on if your GraphQL operation is a [query](#queries), [mutation](#mutations), or a [subscription](#subscriptions). Go to the appropriate API documentation for more information about the functionality and available options for each type.
```
stubailo: This has a reference to subscriptions, but as you mentioned this doesn't have API docs for subscriptions yet right?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+Before we look into the specific behaviors of each operation, let us look at the `config` object.
+
+<h3 id="graphql-config">`config`</h3>
+
+The `config` object is the second argument you pass into the `graphql()` function. After your GraphQL document. The config is optional and allows you to add some custom behavior to your higher order component.
```
stubailo: > The `config` object is the second argument you pass into the `graphql()` function. After your GraphQL document.

Probably should be a comma:

> The `config` object is the second argument you pass into the `graphql()` function, after your GraphQL document.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+  return (
+    <ul>
+      {todos.map(({ id, text }) => (
+        <li key={id}>{text}</li>
+      ))}
+    </ul>
+  );
+}
+
+export default graphql(gql`
```
stubailo: I think we should also demonstrate using an intermediate function, like:

```js
const withTodoAppQuery = graphql(...);
const TodoAppWithData = withTodoAppQuery(TodoApp);
export default TodoAppWithData;
```

I find some people get confused with the multiple function invocations.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+<h3 id="graphql-config.props">`config.props`</h3>
+
+The `config.props` property allows you to define a map function that takes your props including the props added by the `graphql()` function ([`props.data`](#graphql-query-data) for queries and [`props.mutate`](#graphql-mutation-mutate) for mutations) and allows you to compute a new props object that will be provided to the component that `graphql()` is wrapping.
+
+The function you define behaves almost exactly like [`mapProps` from Recompose][] providing the same benefits without the need of another library.
+
+[`mapProps` from Recompose]: https://github.com/acdlite/recompose/blob/2e71fdf4270cc8022a6574aaf00731bfc25dcae6/docs/API.md#mapprops
+
+`config.props` is most useful when you want to abstract away complex functions calls into a simple prop that you can pass down to your component.
```
stubailo: I'd maybe add a note here that this is helpful if you want to make your pure UI components decoupled from GraphQL and Apollo concerns.

source/api.md
calebmer: I your comment may prove my point because the illustration is better written:

```js
funcC(funcB(funcA(component)))
```

As it is the same as:

```js
const x = funcA(component);
const y = funcB(x);
const z = funcC(y);
```

But I see your point. Removed the “non-intuitive” wording.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+    </button>
+  );
+}
+```
+
+<h3 id="graphql-config.skip">`config.skip`</h3>
+
+If `config.skip` is true then all of the React Apollo code will be skipped *entirely*. It will be as if the `graphql()` function were a simple identity function. Your component will behave as if the `graphql()` function were not there at all.
+
+Instead of passing a boolean to `config.skip`, you may also pass a function to `config.skip`. The function will take your components props and should return a boolean. If the boolean returns true then the skip behavior will go into effect.
```
stubailo: I think we should mention this is useful especially if you want to use a different query based on some prop.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+})(MyComponent);
+```
+
+<h3 id="graphql-config.name">`config.name`</h3>
+
+This property allows you to configure the name of the prop that gets passed down to your component. By default if the GraphQL document you pass into `graphql()` is a query then your prop will be named [`data`](#graphql-query-data). If the you pass a mutation then your prop will be named [`mutate`](#graphql-mutation-mutate). While appropriate these default names collide when you are trying to use multiple queries or mutations with the same component. To avoid collisions you may use `config.name` to provide the prop from each query or mutation HOC a new name.
+
+**Example:**
+
+This example uses the [`compose`](#compose) function to use multiple `graphql()` HOCs together.
```
stubailo: I feel like we need some section of the docs called "using multiple mutations or queries on one component" - I guess we have a recipe for that somewhere?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+const MyComponent = graphql(
+  gql`{ ... }`,
+  { withRef: true },
+)(MyWrappedComponent);
+
+class MyContainerComponent extends Component {
+  render() {
+    return (
+      <MyWrappedComponent
```
stubailo: I think you wanted to use `MyComponent` here.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+    );
+  }
+}
+```
+
+<h3 id="graphql-config.alias">`config.alias`</h3>
+
+By default the display name for React Apollo components is `Apollo(${WrappedComponent.displayName})`. This is a pattern used by most React libraries that make use of higher order components. However, it may get a little confusing when you are using more then one higher order components and you look at the [React Devtools][].
+
+[React Devtools]: https://camo.githubusercontent.com/42385f70ef638c48310ce01a675ceceb4d4b84a9/68747470733a2f2f64337676366c703535716a6171632e636c6f756466726f6e742e6e65742f6974656d732f30543361333532443366325330423049314e31662f53637265656e25323053686f74253230323031372d30312d3132253230617425323031362e33372e30302e706e673f582d436c6f75644170702d56697369746f722d49643d626536623231313261633434616130636135386432623562616265373336323626763d3236623964363434
```
stubailo: Where is this image stored?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+        {todos.map(({ id, text }) => (
+          <li key={id}>{text}</li>
+        ))}
+      </ul>
+    );
+  }
+}
+```
+
+<h3 id="graphql-query-data.loading">`props.data.loading`</h3>
```
stubailo: We should also make sure to document networkStatus here.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+}
+```
+
+The `data` prop has some other useful properties which can be accessed directly from `data`. For example, `data.loading` or `data.error`. These properties are documented below.
+
+Make sure to always check `data.loading` and `data.error` in your components before rendering. Properties like `data.todos` which contain your app’s data may be undefined while your component is performing its initial fetch. Checking `data.loading` and `data.error` helps you avoid any issues with undefined data. Such checks may look like:
+
+```js
+render() {
+  const { data: { loading, error, todos } } = this.props;
```
stubailo: I feel like this would be clearer with:

```js
const { loading, error, todos } = this.props.data;
```

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+    );
+  }
+}
+```
+
+<h3 id="graphql-query-data.loading">`props.data.loading`</h3>
+
+A boolean representing whether or not a query request is currently in flight for this component. This means that a query request has been sent using your network interface, and we have not yet gotten a response back. Use this property to render a loading component.
+
+However, just because `data.loading` is true it does not mean that you won’t have data. For instance, if you already have `data.todos`, but you want to get the latest todos from your API `data.loading` might be true, but you will still have the todos from your previous request.
```
stubailo: Is this specifically in the case of `forceFetch` or some non-`cache-only` network policy?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+**Example:**
```
stubailo: I think this example is redundant with the above, but maybe that's fine.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+export default graphql(gql`query { ... }`)(MyComponent);
+```
+
+<h3 id="graphql-query-data.error">`props.data.error`</h3>
+
+If an error occurred then this property will be an instance of [`ApolloError`][]. If you do not handle this error you will get a warning in your console that says something like: `"Unhandled (in react-apollo) Error: ..."`.
+
+[`ApolloError`]: /core/apollo-client-api.html#ApolloError
+
+**Example:**
```
stubailo: Same with this example, also redundant with the above.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+      Query executed with the following variables:
+      <code>{JSON.stringify(variables)}</code>
+    </div>
+  );
+}
+
+export default graphql(gql`query { ... }`)(MyComponent);
+```
+
+<h3 id="graphql-query-data.refetch">`props.data.refetch()`</h3>
```
stubailo: Refetch in the current implementation can optionally take new variables. I take we're intentionally not documenting that? (I would be in support of that, since it's a weird API IMO)

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+**Example:**
+
+```js
+data.fetchMore({
+  updateQuery: (previousResult, { fetchMoreResult }) => {
+    return {
+      ...previousResult,
+      // Add the new feed data to the end of the old feed data.
+      feed: [...previousResult.feed, ...fetchMoreResult.data.feed],
```
stubailo: It's unfortunate that in the current API there's this extra `data` field here, definitely something to fix in a future version.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+<h3 id="graphql-query-data.startPolling">`props.data.startPolling(interval)`</h3>
+
+This function will set up an interval and send a fetch request every time that interval executes. The function takes only one integer argument which allows you to configure how often you want your query to be executed in milliseconds.
+
+Polling is a good way to keep the data in your UI fresh. By refetching your data every 5 seconds (for example) you may effectively emulate realtime data without needing to build up a realtime backend.
+
+If you call `data.startPolling` when your query is already polling then the current polling process will be cancelled and a new process will be started with the interval you specified.
+
+You may also use [`options.pollInterval`](#graphql-query-options.pollInterval) to start polling immediately after your component mounts.
```
stubailo: I think we recommend using the `pollInterval` option if that fits the use case, right?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+**Example:**
+
+```js
+class MyComponent extends Component {
+  componentDidMount() {
+    this.props.data.startPolling(1000);
```
stubailo: This particular example seems to be one where `pollInterval` would be a better way to go.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+  options: (props) => ({
+    variables: {
+      width: props.size,
+      height: props.size,
+    },
+  }),
+})(MyComponent);
+```
+
+<h3 id="graphql-query-options.forceFetch">`config.options.forceFetch`</h3>
```
stubailo: CC @helfer, we should consider documenting only the new fetch policy and pretending like forceFetch and noFetch never existed :]

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+`)(MyComponent);
+```
+
+<h3 id="graphql-mutation-options.refetchQueries">`config.options.refetchQueries`</h3>
+
+Sometimes when you make a mutation you also want to update the data in your queries so that your users may see an up-to-date user interface. There are more fine-grained ways to update the data in your cache which include [`options.updateQueries`](#graphql-mutation-options.updateQueries), and [`options.update`](#graphql-mutation-options.update). However, you can update the data in your cache more reliably at the cost of a higher efficiency by using `options.refetchQueries`.
+
+`options.refetchQueries` will execute one or more queries using your network interface and will then normalize the results of those queries into your cache. Allowing you to potentially refetch queries you had fetched before, or fetch brand new queries.
+
+`options.refetchQueries` is an array of either strings or objects.
```
stubailo: Can you mix both in one array?

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+        variables: {
+          id: props.postID,
+        },
+      },
+    ],
+  }),
+})(MyComponent);
+```
+
+<h3 id="graphql-mutation-options.updateQueries">`config.options.updateQueries`</h3>
```
stubailo: Do we want to de-recommend this in favor of `update`? I would definitely put `update` first since it's the more general API and the one we want people to move towards.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+    },
+  },
+})(MyComponent);
+```
+
+<h3 id="graphql-mutation-options.update">`config.options.update`</h3>
+
+This option allows you to update your store based on your mutation’s result. By default Apollo Client will update all of the overlapping nodes in your store. Anything that shares the same id as returned by the `dataIdFromObject` you defined will be updated with the new fields from your mutation results. However, sometimes this alone is not sufficient. Sometimes you may want to update your cache in a way that is dependent on the data currently in your cache. For these updates you may use an `options.update` function.
+
+`options.update` takes to arguments. The first is an instance of a [`DataProxy`][] object which has some methods which will allow you to interact with the data in your store. The second is the respone from your mutation. Either the optimistic response, or the actual response returned by your server.
```
stubailo: to -> two arguments

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+const query = gql`{ todos { ... } }`
+
+export default graphql(gql`
+  mutation ($text: String!) {
+    createTodo(text: $text) { ... }
+  }
+`, {
+  options: {
+    update: (proxy, { data: { createTodo } }) => {
+      const data = proxy.readQuery({ query });
```
stubailo: I think this is way cleaner than `updateQueries`.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+  graphql(`mutation { ... }`),
+  connect(...),
+)(MyComponent);
+```
+
+<h2 id="server">Server</h2>
+
+`react-apollo` also provides some utilities to aid in server side rendering. To learn how to server side render in your app be sure to read the [recipe for server side rendering](server-side-rendering.html). The following is simply a reference for the APIs of the methods used in server side rendering and not a tutorial teaching you how to set it up.
+
+<h2 id="getDataFromTree">`getDataFromTree(jsx)`</h2>
```
stubailo: Maybe `jsx` -> `component`?

source/api.md
```diff
@@ -0,0 +1,1115 @@
+---
+title: API Reference
+---
+
+<h2 id="gql">``gql`{ ... }` ``</h2>
+
+The `gql` template tag is what you use to define GraphQL queries in your Apollo Client apps. It parses your GraphQL query into the [GraphQL.js AST format][] which may then be consumed by Apollo Client methods. Whenever Apollo Client is asking for a GraphQL query you will always want to wrap it in a `gql` template tag.
+
+You may embed other GraphQL documents inside of other GraphQL documents using template string interpolation. This allows you to use fragments defined in one part of your codebase inside of a query define in a completely different file. See the example below for a demonstration of how this works.
```
calebmer: Really? Didn’t know that.

source/api.md
```diff
@@ -0,0 +1,1115 @@
...
+
+For convenience the `gql` tag is exported from the [`graphql-tag`][] package.
```
calebmer: I changed the wording to:

> For convenience the `gql` tag is re-exported in `react-apollo` from the `graphql-tag` package.

I wanted to get a link in to `graphql-tag`.

source/api.md
calebmer: `foo` gets used in the `query` variable, so an error shouldn’t get thrown, correct?

source/api.md
calebmer: It has a big `TODO` where subscriptions should be 😣

We could just remove references to subscriptions altogether.

source/api.md
calebmer: I don’t think we have an explicit recipe. Do you think the docs for the `compose()` function qualify?

source/api.md
calebmer: It’s stored by GitHub. I pulled it from the PR that added this feature.

source/api.md
calebmer: I like destructuring from `this.props` because it may be pretty for people to miss the `.data` part. Thoughts?

source/api.md
calebmer: Added, what do you think of:

> <h3 id="graphql-query-data.networkStatus">`props.data.networkStatus`</h3>
>
> `data.networkStatus` provides a more detailed view into the state of a network request on your component then [`data.loading`](#graphql-query-data.loading) does. `data.networkStatus` is an enum with different number values between 1 and 8. These number values each represent a different network state.
>
> 1. `loading`: The query has never been run before and the query is now currently running. A query will still have this network status even if a result was returned from the cache, but a query was dispatched anyway.
> 2. `setVariables`: If a query’s variables change and a network request was fired then the network status will be `setVariables` until the result of that query comes back. React users will see this when [`options.variables`](#graphql-query-options-variables) changes on their queries.
> 3. `fetchMore`: Indicates that `fetchMore` was called on this query and that the query created is currently in flight.
> 4. `refetch`: It means that `refetch` was called on a query and the refetch request is currently in flight.
> 5. Unused.
> 6. `poll`: Indicates that a polling query is currently in flight. So for example if you are polling a query every 10 seconds then the network status will switch to `poll` every 10 seconds whenever a poll request has been sent but not resolved.
> 7. `ready`: No request is in flight for this query, and no errors happened. Everything is OK.
> 8. `error`: No request is in flight for this query, but one or more errors were detected.
>
> If the network status is less then 7 then it is equivalent to [`data.loading`](#graphql-query-data.loading) being true. In fact you could replace all of your `data.loading` checks with `data.networkStatus < 7` and you would not see a difference. It is recommended that you use `data.loading`, however.
>
> `data.networkStatus` is useful if you want to display a different loading indicator (or no indicator at all) depending on your network status.
>
> **Example:**
>
> ```js
> function MyComponent({ data: { networkStatus } }) {
>   if (networkStatus === 6) {
>     return <div>Polling!</div>;
>   } else if (networkStatus < 7) {
>     return <div>Loading...</div>;
>   } else {
>     // ...
>   }
> }
>
> export default graphql(gql`query { ... }`)(MyComponent);
> ```

source/api.md
calebmer: This is the case when you have `refetch()`ed a query. You still have the old data, but you are currently loading new data.

source/api.md
calebmer: Yeah, I think example redundancy is fine and even desirable. It enables readers to digest only a section of the API at a time without referring to surrounding context.

source/api.md
calebmer: Yep. Intentionally not documenting that. We don’t support `setVariables()` so we shouldn’t be advertising that you may be able to change variables on `refetch()`.

source/api.md
calebmer: Yeah, I added a note that `options.pollInterval` is recommend above, and I added a comment over this line. We still need an example though 🤷‍♂️

source/api.md
calebmer: Not currently. @helfer’s logic is that since we plan to deprecate the string version it is an anti-goal to mix the two.

source/api.md
calebmer: We could definitely put it first. I feel though as if `options.update` is going to be a little hard for people to move to. At least at first given that it will be tricky to get them access to the right variables. Once we move pagination stuff out of the store I think `options.update` gets a lot more attractive.

calebmer: I moved `update` above `refetchQueries`.

source/api.md
calebmer: Agreed, except for we aren’t addressing the really sticky part in this case, variables 😉

In the real world `todos` is going to be a field with variables and so the user would also have to pass in a variables object which would look like this:

```js
const variables = {
  first: 5,
  after: 'c29tZSBiYXNlNjQgdGhpbmc=',
};

const data = proxy.readQuery({ query, variables });
data.todos.push(createTodo);
proxy.writeQuery({ query, variables, data });
```

This raises a lot of questions. Where do you get these variables? Why are we only changing the data in the store for a specific variable set?

Another option would be client fields or an external data structure for pagination. A client field may look like:

```js
proxy.readQuery({
  query: gql`
    {
      localTodos @client { id text completed }
    }
  `,
});
```

(Which is actually totally supported by the proxy API today if you remove `@client`. This is unintentional, but an interesting side effect)

It is also worth noting that Relay Experimental implements pagination with a client field.

source/api.md
calebmer: I like `jsx` because it’s more descriptive of what you pass in. The API is this:

```js
getDataFromTree(<MyComponent/>);
```

Not:

```js
getDataFromTree(MyComponent);
```

One is JSX and one is the component. The API may also notably be:

```js
getDataFromTree(<p>Hello, <strong>world</strong>!</p>);
```

Which has no components at all! 😊

source/api.md
stubailo: Oh sorry I didn't see that bar was used.

source/api.md
stubailo: Either way works for me, I just find the multi level destructuring weird sometimes.

source/api.md
stubailo: Definitely, but I don't know if people would find them if they were looking for a way to have multiple operations. However we mention that like 4 times in these docs so it's probably easy enough to find.

source/api.md
helfer: I'm going to change networkStatus soon to use the actual strings instead of the numbers. At that point we can update this, ok?

calebmer: Excellent 👍

calebmer:
Looks like CI will fail until someone with `meteor` org access can merge this PR: https://github.com/meteor/hexo-theme-meteor/pull/40

source/api.md
Urigo: I've done another run over the subscriptions docs today.
I suggest that once this PR and the subscriptions PR will be merge I'll do another run over this PR to add/change things that are relevant for subscriptions

calebmer:
Ok, just added some links to the API reference from other documentation articles and I linked from the API reference to other documentation articles. I also deleted any imports and installs of `graphql-tag` and `apollo-client` since we now just import from `react-apollo` (although we do have one import for `toIdValue` which we should also export from `react-apollo`, @helfer). Finally, I removed the HOC article and folded it into the initialization article. It was less valuable after taking away the half-API documentation in it.

source/api.md
```diff
@@ -0,0 +1,1189 @@
+---
+title: API Reference
+---
+
+This is a complete reference of every single feature available in React Apollo. This document is not the best place to start for beginners, but rather a place you should return to whenever you have a question about one of React Apollo’s APIs.
```
helfer: What's the best place to start for beginners? We should link to that. Instead of negative phrasing, maybe something like "If you're just getting started, check out X first"

calebmer: The way I thought about this is it wasn’t the first step in the user’s journey. If you look at the sidebar this comes after the GitHunt API example schema. So I figured if a user was just clicking the “next” button through our sidebar and hit this article they would actually want the *next* relevant article which would be the queries article.

There’s value to the API reference being high up in the sidebar, but not for the beginner who is clicking through the docs so this note is target at that user to get them to click through to the “Queries” article.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+export default graphql(gql`query { ... }`, {
+  noFetch: true,
+})(MyComponent);
+```
+
+<h3 id="graphql-query-options.pollInterval">`config.options.pollInterval`</h3>
+
+The interval in milliseconds at which you want to start polling. Whenever that number of milliseconds elapses your query will be executed using the network interface and another execution will be scheduled using the configured number of milliseconds.
+
+This option will start polling your query immeadiately when the component mounts. If you want to start and stop polling dynamically then you may use [`props.data.stopPolling`](#graphql-query-data.startPolling) and [`props.data.startPolling`](#graphql-query-data.stopPolling).
```
helfer: Alternatively, people can set pollInterval to 0 to stop polling, and anything > 0 to start polling.

helfer: Maybe pollInterval is the only option we should expose?

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+The `mutate` function accepts the same options that [`config.options` for mutations](#graphql-mutation-options) accepts, so to make sure to read through the documentation for that to know what you can pass into the `mutate` function.
+
+The reason the `mutate` function accepts the same options is that it will use the options from [`config.options`](#graphql-mutation-options) _by default_. When you pass an object into the `mutate` function you are just overriding what is already in [`config.options`](#graphql-mutation-options).
+
+**Example:**
+
+```js
+function MyComponent({ mutate }) {
+  return (
+    <button onClick={() => mutate()}>
```
helfer: I think this example should be a bit more complete and also include variables (it's very rare to call a mutation without variables).

helfer: Since it's in the later examples, maybe we can just not put this example here? Or alternatively, maybe the initial example should show off the most common use-case.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+      foo: props.foo,
+      bar: props.bar,
+    },
+  }),
+})(MyComponent);
+```
+
+<h3 id="graphql-mutation-options.optimisticResponse">`config.options.optimisticResponse`</h3>
+
+Often when you mutate data it is fairly easy to predict what the response of the mutation will be without asking your server. The optimistic response option allows you to make your mutations feel faster by simulating the result of your mutation in your UI before the mutation actually finishes.
```
helfer: without -> before.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+      data.todos.push(createTodo);
+      proxy.writeQuery({ query, data });
+    },
+  },
+})(MyComponent);
+```
+
+<h3 id="graphql-mutation-options.refetchQueries">`config.options.refetchQueries`</h3>
+
+Sometimes when you make a mutation you also want to update the data in your queries so that your users may see an up-to-date user interface. There are more fine-grained ways to update the data in your cache which include [`options.updateQueries`](#graphql-mutation-options.updateQueries), and [`options.update`](#graphql-mutation-options.update). However, you can update the data in your cache more reliably at the cost of a higher efficiency by using `options.refetchQueries`.
```
helfer: at the cost of a higher efficiency -> at the cost of efficiency

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+
+`options.refetchQueries` will execute one or more queries using your network interface and will then normalize the results of those queries into your cache. Allowing you to potentially refetch queries you had fetched before, or fetch brand new queries.
+
+`options.refetchQueries` is an array of either strings or objects.
+
+If `options.refetchQueries` is an array of strings then Apollo Client will look for any queries with the same names as the provided strings and will refetch those queries with their current variables. So for example if you have a GraphQL query component with a query named `Comments` (the query may look like: `query Comments { ... }`), and you pass an array of strings containing `Comments` to `options.refetchQueries` then the `Comments` query will be re-executed and when it resolves the latest data will be reflected in your UI.
```
helfer: We should no longer document this. We want to discourage people from using it.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+
+The `<ApolloProvider/>` component takes the following props:
+
+- `client`: The required [`ApolloClient`][] instance. This [`ApolloClient`][] instance will be used by all of your components enhanced with GraphQL capabilties.
+- `[store]`: This is an optional instance of a Redux store. If you choose to pass in your Redux store here then `<ApolloProvider/>` will also provide your Redux store like the [`react-redux` `<Provider/>` component][]. This means you only need to use one provider component instead of two!
+
+[`react-redux` `<Provider/>` component]: https://github.com/reactjs/react-redux/blob/master/docs/api.md#provider-store
+[`ApolloClient`]: ../core/apollo-client-api.html#apollo-client
+
+**Example:**
```
helfer: Is it worth showing how you could access the client directly inside your components?

source/api.md
calebmer: I don’t actually think that people should use `networkStatus` over loading. It’s a little more complicated then what you need in the 90% case. I’ll add a note anyway though.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+You may embed a GraphQL document containing only fragments inside of another GraphQL document using template string interpolation. This allows you to use fragments defined in one part of your codebase inside of a query define in a completely different file. See the example below for a demonstration of how this works.
+
+For convenience the `gql` tag is re-exported in `react-apollo` from the [`graphql-tag`][] package.
+
+[GraphQL.js AST format]: https://github.com/graphql/graphql-js/blob/d92dd9883b76e54babf2b0ffccdab838f04fc46c/src/language/ast.js
+[`graphql-tag`]: https://www.npmjs.com/package/graphql-tag
+
+**Example:**
+
+Notice how in the `query` variable we not only include the `fragments` variable through template string interpolation (`${fragments}`), but we also include a spread for the `foo` fragment in our query. If we were to pass the `query` variable to many of our Apollo Client methods it would attempt to execute that query.
```
calebmer: I just deleted the last sentence. It isn’t really necessary.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+const query = gql`
+  query {
+    ...foo
+  }
+
+  ${fragments}
+`;
+```
+
+<h2 id="graphql">`graphql(query, config)(component)`</h2>
```
calebmer: Ok. The reason it wasn’t first is this section makes up more then half of the docs so the `gql` tag would be way further down. I don’t think that’s as big as an issue as I initially thought, however.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+
+Instead of passing a boolean to `config.skip`, you may also pass a function to `config.skip`. The function will take your components props and should return a boolean. If the boolean returns true then the skip behavior will go into effect.
+
+`config.skip` is especially useful if you want to use a different query based on some prop. You can see this in an example below.
+
+**Example:**
+
+```js
+export default graphql(gql`{ ... }`, {
+  skip: true,
```
calebmer: The `skip` option?

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+  console.log(props.updateTodo);
+  console.log(props.deleteTodo);
+
+  return null;
+}
+```
+
+<h3 id="graphql-config.withRef">`config.withRef`</h3>
+
+By setting `config.withRef` to true you will be able to get the instance of your wrapped component from your higher-order GraphQL component using a `getWrappedInstance` method available on the instance of your higher-order GraphQL component.
```
calebmer: Done. I also updated the example to show when it is useful.

* * *

Fun story, in PostGraphQL we *set the state* on a component instance that we have access to via `ref` in GraphiQL: https://github.com/calebmer/postgraphql/blob/6cf9ac7d1b28440523d6938e0c993d007f0169ed/src/postgraphql/graphiql/src/components/PostGraphiQL.js#L175-L180

So, one use case for refs is breaking all paradigms of encapsulation that React likes to sell 😊

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+    </div>
+  );
+}
+
+export default graphql(gql`query { ... }`)(MyComponent);
+```
+
+<h3 id="graphql-query-data.refetch">`props.data.refetch()`</h3>
+
+Forces your component to re-execute the query you defined in the `graphql()` function. This method is helpful when you want to reload the data in your component, or retry a fetch after an error.
```
calebmer: Well if the user has a local schema then there will be no request. I use the word “execute” (in any context) to mean a GraphQL document that Apollo has sent to the user to be executed against their GraphQL operation resolver wherever it may be.

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+- `[variables]`: The optional variables you may provide that will be used with either the `query` option or the query from your `graphql()` HOC (depending on whether or not you specified a `query`).
+- `updateQuery(previousResult, { fetchMoreResult, queryVariables })`: This is the required function you define that will actually update your paginated list. The first argument, `previousResult`, will be the previous data returned by the query you defined in your `graphql()` function. The second argument is an object with two properties, `fetchMoreResult` and `queryVariables`. `fetchMoreResult` is the data returned by the new fetch that used the `query` and `variables` options from `data.fetchMore`. `queryVariables` are the variables that were used when fetching more data. Using these arguments you should return a new data object with the same shape as the GraphQL query you defined in your `graphql()` function. See an example of this below, and also make sure to read the [pagination](pagination.html) recipe which has a full example.
```
calebmer: Such as?

source/api.md
```diff
@@ -0,0 +1,1189 @@
...
+    // ...
+  }
+}
+
+export default graphql(gql`query { ... }`)(MyComponent);
+```
+
+<h3 id="graphql-query-data.stopPolling">`props.data.stopPolling()`</h3>
+
+By calling this function you will stop any current polling process. Your query will not start polling again until you call `props.data.startPolling`.
```
calebmer: I don’t recall seeing `setOptions` in the type definitions for `react-apollo`…

source/api.md
calebmer: We should expose `startPolling` if we are going to expose `stopPolling` and there is good reason to expose `stopPolling`.

source/api.md
calebmer: I added an example variables object. I think of examples in this context as abstract illustrations of what this specific part of the API looks like when used.

source/api.md
calebmer: I figure we should remove this at the same time we remove the docs for `updateQueries`. In the meantime we still don’t have a story around updating the specific variables being rendered right now by your components whereas string `refetchQueries` do provide that capability.

calebmer:
Made your recommended changes @helfer and added docs on `fetchPolicy` (which is worth reviewing). One note is that we will still need a PR to the core docs to add information about `fetchPolicy` 👍

source/api.md
helfer: Ah, yeah, for react-apollo that's the same as simply changing the options via props, but you can only reasonably do that from a parent component.

calebmer:
Note that this does have some `TODO`s in the docs around subscriptions.

PR #168, opened 2017-03-04
Update prefetching.md

developer239:
To ensure that the query gets fired only once.


helfer:
Thanks @developer239. Is the `onMouseEnter` event supported by all major browsers these days? And have you actually observed queries being fired multiple times?

developer239:
Ha! I will post here video tomorrow I had a problem with queries being fired multiple times. I am not sure though if onMouseEnter is supported in all modern browsers though.

developer239:
When I put `onMouseOver` event in `<Row>` component it bubbled and fired multiple times.

https://github.com/developer239/webpack-doc/blob/master/src/modules/blog/components/List/index.jsx#L19

I know that it is probably not a really good idea putting prefetching on such a large element. However I believe that the query shout be fired only once. `onMouseEnter` does not solve the problem because it is not supported by all major browsers.

PR #172, opened 2017-03-10
Add creating an ApolloClient with mock schema

magbicaleman:
Want to add an example of working with mock schema, when a developer is reading about creating a client.

### DIFF
--- a/source/initialization.md
+++ b/source/initialization.md
@@ -37,6 +37,27 @@ const client = new ApolloClient({
 
 `ApolloClient` has some other options which control the behavior of the client, and we'll see examples of their use throughout this guide.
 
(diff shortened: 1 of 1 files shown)


stubailo:
I don't think the intialization page is the right place - perhaps we can add a new recipe page, "mocking"?

magbicaleman:
@stubailo Personally I would have found it very useful if it was in the "Create a Clinet" section, but I definitely understand what your saying. 

So your recommendation is to create a new "Mocking" page under "RECIPES". 

I think at the very least it would be useful to have a mention of it in the `Setup and options`, what do you think?

helfer:
I agree with @stubailo. Let's make a "mocking" section under recipes @magbicaleman. It's probably best to open a new PR for that, so I'll close this one.

PS: sorry for the long wait.

PR #176, opened 2017-03-15
Use `reduxRootSelector` rather than `reduxRootKey`

grifotv:
as per `apollo-client` version `v1.0.0-rc.2`


grifotv:
https://github.com/apollographql/apollo-client/pull/1401

grifotv:
Correct approach https://github.com/apollographql/react-docs/pull/179

PR #180, opened 2017-03-20
Clarify what is "Query" key in `customResolvers` example

nosovsh:
Currently it is not clear what is it. Is it a special keyword or some magic or something else.
In my case it was "Root" and I spent pretty much time trying to understand why `customResolvers` are not working for me.

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -106,7 +106,7 @@ const client = new ApolloClient({
     });
 ```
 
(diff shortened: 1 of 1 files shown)


stubailo:
In fact, it can be any type name! Might be worth adding that as well.

helfer:
Thanks for the PR @nosovsh, I've merged it as-is.

PR #181, opened 2017-03-20
Update `customResolvers` docs

nosovsh:
Added that
1. it is possible to return list of ids #508 https://github.com/apollographql/react-apollo/issues/508
2. it is possible to dynamically skip cache by returning `toIdValue(null)` or `[toIdValue(null)]`. Actually I'm not sure that it is legal way to do it, but it works :)

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -108,6 +108,16 @@ const client = new ApolloClient({
 
 Apollo Client will use the return value of the custom resolver to look up the item in its cache. `toIdValue` must be used to indicate that the value returned should be interpreted as an id, and not as a scalar value or an object.
 
+It is possible to return a list of ids if you expect a list of entities.
+```
+customResolvers: {
+  Query: {
+    books: (_, args) => args['ids'].map(id => toIdValue(dataIdFromObject({ __typename: 'book', id: id }))),
+  },
+},
+```
+Also it is possibly to dynamically decide inside resolver whether you want to use cache for this request or not. If not just return `toIdValue(null)` or `[toIdValue(null)]` for list fields.
+
 <h2 id="fetchMore">Using `fetchMore`</h2>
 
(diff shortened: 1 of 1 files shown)


source/cache-updates.md
```diff
@@ -108,6 +108,16 @@ const client = new ApolloClient({
...
 
+It is possible to return a list of ids if you expect a list of entities.
+```
+customResolvers: {
+  Query: {
+    books: (_, args) => args['ids'].map(id => toIdValue(dataIdFromObject({ __typename: 'book', id: id }))),
+  },
+},
+```
+Also it is possibly to dynamically decide inside resolver whether you want to use cache for this request or not. If not just return `toIdValue(null)` or `[toIdValue(null)]` for list fields.
```
stubailo: Hmm - is this because the store reading looks inside and decides the object is not in the store? This definitely seems sketchy, what's the use case where you want to do this?

helfer: Yeah, I don't think we should recommend this to people. I'm interested in hearing your use-case though.

helfer:
Thanks for the PR. I've decided to remove the part about forcing cache misses with resolvers. I don't think we should document that, but I'm interested in your use-case.

nosovsh:
@helfer @stubailo Thanks for reviewing and merging! 
My use case for skipping cache:

Let's say I have query `users` that accepts different variables like:
```
{
  ids: ['1','2'],
  search_str: 'find me',
  sort: 'data.name',
  ...
}
```
When query is executed with only `ids` in variables I can precisely say what ids should be returned and in which order and I can help Apollo to find them in cache.
But if query is executed with other variables I have no idea what server will return so I just skipping `customResolvers` optimization.

stubailo:
Have you tried just returning `null` from the resolver? I don't know why `toIdValue` needs to be involved.

nosovsh:
If is working if you are returning `null` for query that expects single entity in result (it will bypass cache) but you will get exception if you return it for list queries.

PR #182, opened 2017-03-20
Typo in Pagination docs

alfonsoperez:
Fix first arguments in graphql function call in Pagination docs

### DIFF
--- a/source/pagination.md
+++ b/source/pagination.md
@@ -125,7 +125,7 @@ const MoreCommentsQuery = gql`
   }
 `;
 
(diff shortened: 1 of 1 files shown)


stubailo:
This is intentional - the GraphQL HoC is used with the original query, and in `fetchMore` we use a _different_ query to fetch more data.

I'd be happy to accept a PR making that clearer, but this is not a typo.

alfonsoperez:
Gotcha, I didn't see that first Query defined so I assumed wrongly. I will take some time in the weekend to see if I can help in this regard. Thanks for the clarification.

PR #184, opened 2017-03-22
Update subscribeToMore specifying the updateQuery option as optional

cesarsolorzano:
apollographql/apollo-client#1455


stubailo:
Great, thanks!

PR #188, opened 2017-03-29
Add note for options.notifyOnNetworkStatusChange

ligaz:
Add a small note about using `option.notifyOnNetworkStatusChange` in the `query` function.


stubailo:
Thanks a lot!

PR #198, opened 2017-04-03
Sanitize JSON.stringify output to prevent XSS

vlasenko:
See these docs for rationale:
http://redux.js.org/docs/recipes/ServerRendering.html#security-considerations|


stubailo:
This is a great call, thanks for the PR!

PR #204, opened 2017-04-07
docs(api): Add docs for variables argument in docs of data.refetch

huang475:
Not sure should update the example though.


helfer:
Thanks @huang475! Let me know if you want to get more involved in the project, we're actively looking for new contributors.

PR #208, opened 2017-04-12
Fix links to server side rendering recipe

ewdicus:
Links to `server-side-rendering.html` had an extra `#` in front.


helfer:
Good catch @ewdicus !

PR #209, opened 2017-04-13
Adds a minimal mention of Direct Cache Access

Siyfion:
And links to the relevant bit in the Vanilla JS section.

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -290,6 +290,10 @@ The main difference between the two is where the update behavior is declared. Wi
 
 We recommend using the `reducer` option, except when there's a good reason to use `updateQueries` instead (eg. if it would make your app much easier to understand and maintain).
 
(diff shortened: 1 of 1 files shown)


helfer:
Thanks for the PR @Siyfion! Can you make the example a full one that includes decorating a component with `graphql` and then using the `mutate` function inside of it? That would be awesome!

stubailo:
I think this is actually the best documentation to link to maybe? http://dev.apollodata.com/react/api-mutations.html#graphql-mutation-options-update

Siyfion:
@stubailo & @helfer I've changed the link to the one Sashko posted and I've fleshed out the example a bit more. I can expand it even more, but I didn't want to write an _enormous_ example in comparison to the others on the page, as it might be a bit overkill? Thoughts please!

stubailo:
This is great, thanks!

Siyfion:
Always happy to help. 👍

PR #210, opened 2017-04-17
Fix SSR link

ooade:
* Fix SSR link


helfer:
Oops! 😀 

Thanks for the PR @ooade !

PR #211, opened 2017-04-17
Use fetchPolicy for legacy forceFetch

glennreyes:


helfer:
Hey @glennreyes, thanks for the PR! forceFetch corresponds to network-only, so I've changed it.

PR #216, opened 2017-04-30
Adds documentation for partial data results

n1ru4l:


source/partial-data.md
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type Query {
+  series: [Series]!
+  oneSeries(id: Int!): Series!
+  episodes(seriesId: Int!): [Episode]!
+}
+```
+
+And you have two Scenes:
```
stubailo: I don't think these are usually called scenes in react - maybe views or routes?

source/partial-data.md
```diff
@@ -0,0 +1,167 @@
+---
+title: Displaying Partial Data
```
stubailo: I'd probably call this "Displaying preview data" or something, since it's not really partial data? That's what the feature used to be called but not anymore.

source/partial-data.md
```diff
@@ -0,0 +1,167 @@
...
+query seriesEpisodes($seriesId: Int!) {
+  episodes(seriesId: $seriesId) {
+    id
+    title
+    cover
+  }
+}
+```
+
+By adding a customResolver for the `oneSeries` field (and having dataIdFromObject function which normalizes the cache), the data can be resolved instantly from the store without a server round trip.
```
stubailo: Probably good to link to custom resolvers in the docs here: http://dev.apollodata.com/react/cache-updates.html#cacheRedirect

source/partial-data.md
```diff
@@ -0,0 +1,167 @@
...
+
+SeriesDetailSceneWithData.propTypes = {
+  seriesId: PropTypes.number.isRequired,
+}
+
+export default SeriesDetailScene
+
+```
+
+Unfortunately if the user would now visit the second scene without ever visiting the first scene this would result in two network requests (since the data for the first query is not in the store yet). By using a `BatchedNetworkInterface` those two queries can be send to the server in one network request.
```
stubailo: Couldn't this be fixed by setting a `cache-only` network policy on the smaller query?

stubailo:
Thanks for the PR! Some comments in the review.

source/partial-data.md
n1ru4l: Maybe "Query Splitting" or "Improve UI By Splitting Queries"? I also think "Partial Data Results" is misleading but "Displaying preview data" is also not very meaningful.

source/partial-data.md
n1ru4l: Depends on what libs you use for navigation. In navigator for react-native they call the views scenes. But I also think that view is more generic.

source/partial-data.md
n1ru4l: Wouldn't that mean that the query is never fetched?

n1ru4l: Or do you mean that the second query should also request the data of the first query?

n1ru4l:
bump! I added some commits and added some questions to your annotations :)

source/query-splitting.md
```diff
@@ -0,0 +1,167 @@
...
+    title
+    description
+    cover
+  }
+}
+```
+
+```graphql
+query seriesEpisodes($seriesId: Int!) {
+  episodes(seriesId: $seriesId) {
```
helfer: I think it would be nicer to change this query to: ```
{
  series(id: $seriesId) {
    id
    episodes {
      id
      title
      cover
    }
  }
}
```

That way you don't have to change the schema on the server at all to accommodate for splitting the queries on the client.

helfer:
@n1ru4l thanks a lot for this PR! I added one more comment. Let me know what you think about it, maybe change it, and then we'll merge it! 👍

source/query-splitting.md
helfer: I just think it's not idiomatic graphql to have a separate `episodes` root field. It looks more like a RESTful API that way.

n1ru4l: Just realized my statement does not apply to your comment at all :sweat:

Makes totally sense to me, I will update it!

source/partial-data.md
n1ru4l: @helfer @stubailo Is it possible to handle the error thrown by a `cache-only` query? If so this would be a really cool addition!

n1ru4l:
@helfer I applied your suggestions. I also added a comment to a previous annotation by @stubalio (Would be awesome if you could give me hints on this, otherwise I will try some experiments at the end of week).

PR #232, opened 2017-05-25
Update auth.md - add info about "withApollo"

caspg:
I think it will be really helpful to mention that `withApollo` HOC is needed to access `client`.


stubailo:
Thanks!

PR #235, opened 2017-06-02
subscribeToMore returns unsubscribe handler

alexicum:
>@deep-c It's not defined because you're not actually getting an observable back. Apollo Client subscribes to it for you, so all you get back is actually the unsubscribe handle. So if you call this.subscription() it will unsubscribe. Maybe the variable name I chose in GitHunt-React isn't great 😄 . Would be better to call it unsubscribeHandle or something.

https://github.com/apollographql/apollo-client/issues/1145#issuecomment-271566605


helfer:
Thanks for the PR @alexicum !

PR #238, opened 2017-06-04
Fixed faulty import for IntrospectionFragmentMatcher

timbrandin:


stubailo:
`react-apollo` in recent versions re-exports everything from `apollo-client`, so this should actually work. Doesn't it? What versions are you using?

timbrandin:
Oh, I see, maybe we're not using the latest version then. We're on 0.13.3 of react-apollo and 1.3.0 of apollo-client.

stubailo:
Yeah, `react-apollo` is at `1.4.1` now - might be worth looking at upgrading!

stubailo:
Since we don't currently do documentation versioning, I'm going to close this. Probably around a 2.0 release we should start versioning docs.

PR #240, opened 2017-06-06
Document the expected updateQuery return value

ArnaudRinquin:
By specifying that it should have the same shape as the original query data.


stubailo:
Thanks!

PR #241, opened 2017-06-13
Fix missing variables in docs for refetchQueries

TimMikeladze:


helfer:
Thanks @TimMikeladze !

source/cache-updates.md
```diff
@@ -92,7 +92,7 @@ mutate({
   //... insert comment mutation
   refetchQueries: [{
     query: gql`
-      query updateCache {
+      query updateCache($repoName: String!) {
```
stubailo: This needs to be `repoFullName` to match the key in `variables:`

helfer: Then it will be inconsistent with the variable name used in the query. Something's got to give. I guess we just forgot to rename the second one?

PR #242, opened 2017-06-16
Update initialization.md to add example build step

ngbrown:
Added example for building the schema automatically for the IntrospectionFragmentMatcher.  This example uses Jake, but it's straightforward to use in any build system.


stubailo:
I think this is too big to include in the critical path of the initialization docs. Can you create a new page under "Recipes", and then link to it here instead please? That way, we can also add examples with different systems without making this page longer and longer.

stubailo:
Ping, this would be great to have!

ngbrown:
I think it probably isn't enough for an entire recipe.  What if I make it more generic, just a node.js function, and use it to replace the  three steps in [Using Fragments on unions and interfaces](http://dev.apollodata.com/react/initialization.html#fragment-matcher)?  That way the page won't grow much and there won't be an option per build system.

stubailo:
I think it is definitely enough for a new page - we can make it about the fragment matcher in general? Or it can be a page about build tooling and we can later add something about how to use `persistgraphql` or `apollo-codegen`!

peggyrayzis:
Thanks for your contribution @ngbrown! Although we aren't able to merge this PR, your contribution definitely helped us update our fragment matcher section of the docs to include the build step. Ultimately, we decided to keep it simple and use a node script instead of using Jake. 😀

ngbrown:
@peggyrayzis thanks for cleaning it up and including it in the documentation.

PR #243, opened 2017-06-20
Change the endpoint URI example to the less misleading one

abalone0204:
People don't usually use `http://localhost:3000/` as their endpoint for graphql,
I think it might be more understandable what it is if we use `http://localhost:3000/graphql` as example.


stubailo:
Thanks, this is a good improvement

PR #246, opened 2017-06-28
Fix broken `config.option` links

neilff:
I noticed a couple of the anchor links were not pointing to their correct URLs, this should fix the problem! I _believe_ the second commit also fixes the broken generic `config.options` link, but I wasn't 100% sure if it was pointing to the correct reference.


helfer:
Thanks @neilff !

PR #247, opened 2017-06-29
Add documentation for the new connection directive

shadaj:


source/cache-updates.md
```diff
@@ -410,6 +410,25 @@ Here, the `fetchMore` query is the same as the query associated with the compone
 
 Although `fetchMore` is often used for pagination, there are many other cases in which it is applicable. For example, suppose you have a list of items (say, a collaborative todo list) and you have a way to fetch items that have been updated after a certain time. Then, you don't have to refetch the whole todo list to get updates: you can just incorporate the newly added items with `fetchMore`, as long as your `updateQuery` function correctly merges the new results.
 
+<h3 id="connection-directive">The `@connection` directive</h3>
+By default, the result of a `fetchMore` will be stored in the cache according to the query run and its parameters. Because parameters usually change between calls to `fetchMore`, this means that the results of each `fetchMore` will be placed in a different location in the store. This can be confusing when using imperative store updates, as there is no stable location in the cache for `fetchMore` results.
```
helfer: This is technically true, but it's not a behavior people should rely on. For the purpose of pagination, the `updateQuery` that runs during a `fetchMore` actually stores everything to the field of the initial query with the same parameters. If the variables aren't always the same it can thus be hard to know the store location for an `update` somewhere else.

source/pagination.md
```diff
@@ -222,3 +222,20 @@ const CommentsWithData = graphql(CommentsQuery, {
   },
 })(Feed);
 ```
+
+<h2 id="connection-directive">The `@connection` directive</h2>
+When using paginated queries, results from individual queries can become scattered in the store, as query parameters are used to determine the default store key but often change with pagination. This is problematic for imperative store updates, as there is no stable store key for updates to target. To direct Apollo Client to use a stable store key for paginated queries, you can use the `@connection` directive to specify a store key for parts of your queries. For example, if we wanted to have a stable store key for the feed query earlier, we could adjust our query to use the `@connection` directive:
```
helfer: As mentioned above, they're actually stored in the same location, but that location isn't always easy to know.

source/pagination.md
```diff
@@ -222,3 +222,20 @@ const CommentsWithData = graphql(CommentsQuery, {
...
+  query Feed($type: FeedType!, $offset: Int, $limit: Int) {
+    currentUser {
+      login
+    }
+    feed(type: $type, offset: $offset, limit: $limit) @connection(key: "feed") {
+      id
+      # ...
+    }
+  }
+`;```
```
helfer: The highlighting here is weird, I think it's because the three backticks aren't on their own line.

source/cache-updates.md
```diff
@@ -410,6 +410,25 @@ Here, the `fetchMore` query is the same as the query associated with the compone
...
+    login
+  }
+  
+  feed(type: $type, offset: $offset, limit: $limit) @connection(key: "feed", filter: ["type"]) {
+    ...FeedEntry
+  }
+}
+```
+
+With the above query, even with multiple `fetchMore`s, the results of each feed update will always result in the `feed` key in the store being updated with the latest accumulated values. In this example, we also use the `@connection` directive's optional `filter` argument to include the `type` query argument in the store key, which results in multiple store values that accumulate queries from each type of feed.
```
stubailo: Can we add an example of using `writeQuery` to update the above example? This shows how to use the directive inside `fetchMore`, but not how to update the data, which is the main use case.

source/cache-updates.md
```diff
@@ -410,6 +410,25 @@ Here, the `fetchMore` query is the same as the query associated with the compone
...
 
+<h3 id="connection-directive">The `@connection` directive</h3>
+By default, the result of a `fetchMore` will be stored in the cache according to the initial query executed and its parameters. Due to this behavior, it can be hard to know the location in the cache to run an imperative update on if the variables from the initial query are not known, which often happens if we are running store updates from a different place than where the queries are executed.
+
+To have a stable cache location for query results, Apollo Client 1.6 introduced the `@connection` directive, which can be used to specify a custom store key for results. To use the `@connection` directive, simply add the directive to the segment of the query you want a custom store key for and provide the `key` parameter to specify the store key. In addition to the `key` parameter, you can also include the optional `filter` parameter, which takes an array of query argument names to include in the generated custom store key.
+
+```
+query query Feed($type: FeedType!, $offset: Int, $limit: Int) {
```
stubailo: `query query`

source/pagination.md
```diff
@@ -222,3 +222,21 @@ const CommentsWithData = graphql(CommentsQuery, {
...
+    }
+    feed(type: $type, offset: $offset, limit: $limit) @connection(key: "feed", filter: ["type"]) {
+      id
+      # ...
+    }
+  }
+`;
+```
+
+This would result in the accumulated feed in every query or `fetchMore` being placed in the store under the `feed` key, which we could later use of imperative store updates. In this example, we also use the `@connection` directive's optional `filter` argument, which allows us to include some arguments of the query in the store key. In this case, we want to include the `type` query argument in the store key, which results in multiple store values that accumulate pages from each type of feed.
```
stubailo: I'd add something in this section about the connection directive being optional.

shadaj:
@stubailo ready for another review!

stubailo:
Alright, it's in!

PR #250, opened 2017-07-03
idFromObject should optimally have typename

blocka:


stubailo:
The default one already uses that! This is in the section:

>  You can also specify a custom function to generate IDs from each object

So it's specifically trying to use an unconventional approach.

Happy to accept a different PR to make the example clearer.

PR #251, opened 2017-07-04
Update api-mutations.md

PDS42:
l.204  you are declaring data as a `const`, then pushing values to the variables l.205, I changed `const` to `let`.


stubailo:
You can push to a const array/object.

```js
const x = [];
x.push(5);
console.log(x); // [5]
```

Const just stops the reference from changing. So you can't do:

```js
const x = [];
x = x.concat([5]); // ERROR
```

PR #260, opened 2017-07-22
Improve customResolvers documentation

mxstbr:
Add note about having to be the same data, add note about where to get `__typename`, improve overall documentation.

> Sidenote: I feel like `customResolvers` isn't a super great name since it's not like it replaces a resolver on the backend, it helps apollo resolve something from the client cache. What about renaming it to `cacheResolvers` instead?

### DIFF
--- a/source/cache-updates.md
+++ b/source/cache-updates.md
@@ -473,6 +473,8 @@ query DetailView {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


stubailo:
Thank you @mxstbr!!

PR #266, opened 2017-08-06
Remove wording about "just like above"

blargity:
The previous example is actually only implementing `optimisticResponse` but then the next example implements `update` as well as `optimisticResponse`, and I found the wording to be a bit confusing around this. Do we need to implement `update` as well as `optimisticResponse` as the example shows for adding to a list? If so, the wording was incorrect. If not, then it'd be ideal to give an example where only `optimisticResponse` is required for adding to a list.

I'm currently assuming that the wording was actually slightly off, and that `update` was in fact necessary, in which case we can just remove the bit about `update` not being required.


stubailo:
Yeah this is a good change - you need both.

PR #267, opened 2017-08-06
fix example; cursor is not available on root

koenpunt:
The example suggested that `cursor` would be returned alongside `moreComments`, but in reality the cursor is returned as a field of `moreComments`.

I also updated some ambiguous naming


stubailo:
Thank you!

PR #270, opened 2017-08-09
Add link to FuseBox plugin

otothea:
I have been using apollo-client and webpack on a project, but we are switching to use the FuseBox bundler and had to create a plugin to work with your webpack loader. I saw the link for the Jest loader and thought I might share the FuseBox loader here as well.


stubailo:
That's awesome, thanks! The CLA is just to verify that you assign us the license of the text you've added to this documentation site, not of your tool or anything like that, so it would be great if you could sign it!

stubailo:
Thanks!

PR #273, opened 2017-08-16
Add gql to the list of imports

locnguyen:
gql is used to create the graphql query for the Profile component so it has to be imported to work.


stubailo:
Awesome, thanks!

PR #277, opened 2017-08-18
update docs: PropTypes is now a separate package

staylor:


stubailo:
Thanks for this one as well!

PR #281, opened 2017-08-23
Fixed small typo in example

robinvdvleuten:
In found a small typo in the example while reading the docs.


stubailo:
Both are correct actually!

stubailo:
But it looks nicer the new way

PR #283, opened 2017-08-25
Setup a build step to generate fragment schema

Zakjholt:


peggyrayzis:
Thanks for your contribution @Zakjholt! Although we aren't able to merge this PR, your contribution definitely helped us update our fragment matcher section of the docs to include the build step. Ultimately, we decided to go with a node script instead of using curl. 😀

Zakjholt:
No problem! A node script probably makes a bit more sense ^.^

PR #287, opened 2017-09-07
[RFR] Update SSR section again

eschaefer:
Removes note which referenced previous version of GitHunt-React app.


stubailo:
@eschaefer oh, awesome!

PR #292, opened 2017-09-19
Add docs for graphql-tag precompilation with babel

leoasis:
This adds an entry in the recipes to explain how to precompile the queries with Babel using two approaches: [babel-plugin-graphql-tag](https://github.com/gajus/babel-plugin-graphql-tag) and [graphql-tag.macro](https://github.com/leoasis/graphql-tag.macro).

In some unrelated news, had some problems trying to see the documentation locally, did `npm start` and tried to access the url, but I just saw a blank page and had this warning in the console:

```
INFO  Start processing
INFO  Hexo is running at http://localhost:4000/react/. Press Ctrl+C to stop.
WARN  No layout: index.html
```

Also, all generated files with `npm run generate` where empty html files.

Not sure what I'm doing wrong, so any help with this is appreciated 😄


stubailo:
@leoasis it's possible that you didn't initialize the submodule, so you might need to do a `git submodule init; git submodule update` to get the theme.

source/babel.md
```diff
@@ -0,0 +1,99 @@
+---
+title: Precompilation with babel
+---
+
+If you prefer co-locating your GraphQL queries in your Javascript files, you typically use the [graphql-tag](https://github.com/apollographql/graphql-tag) library to write them. That requires to process the query strings into a GraphQL AST, which will add to the startup time of your application, especially if you have many queries.
```
stubailo: > which will add to the startup time of your application, especially if you have many queries.

Do we have any numbers on this? I'd say the main reason is likely to avoid the bundle size of loading the parser.

leoasis: I created a repo to have some numbers on this: https://github.com/leoasis/perf-graphql-tag. It sets up 400 `gql` calls (200 fragments and 200 queries including a fragment). It generates two files, one using graphql-tag, and another one preprocessed with babel with the `babel-plugin-graphql-tag` plugin.

In my computer the numbers show ~85ms for the runtime version, and ~20ms for the preprocessed version. 

With 400 fragments and 400 queries, I get these numbers: ~157ms runtime, ~28ms preprocessed.

So it seems the more queries you have, the more time preprocessing saves.

Maybe we could put both benefits in the docs?

PR #293, opened 2017-09-22
Add rootEl to redux.md

iamclaytonray:
In your DOM rendering, you threw in rootEl, but rootEl was never defined. Assuming the ID is "root", I added a const that pointed to that div/element.

`const rootEl = document.querySelector("#root");`


kavink:
is the redux doc still valid with 2.0 release ? Can we still use redux along with new in-memory cache ?

iamclaytonray:
^ Try asking questions on SO or browse the issues

PR #301, opened 2017-10-11
Fix API reference link of queries.md

tsh3lley:
api-queries.html is a broken link, api-queries.md is the correct url


jbaxleyiii:
@tsh3lley the original should actually work on the docs site?

tsh3lley:
@jbaxleyiii whoops - correct.

PR #303, opened 2017-10-19
Fix broken Githunt example links

joerodrig:
URLs using `/ui` were updated to use `/src`.

The URL for files referenced changed in this commit:
https://github.com/apollographql/GitHunt-React/commit/91c3f259af201834254b42143d977f82588c5c3f#diff-935c31f6d60f89156e3f8f203939006e

### DIFF
--- a/source/fragments.md
+++ b/source/fragments.md
@@ -78,7 +78,7 @@ export const COMMENT_QUERY = gql`
 `;
 ```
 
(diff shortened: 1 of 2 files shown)


stubailo:
Hi, thanks for the PR - the docs have been moved to the same repository as the code for 2.x. I don't know if we made this change there, but we are working on improving GitHunt significantly.

PR #304, opened 2017-10-19
Update link in mutations.md

petetnt:
https://github.com/apollographql/react-docs/commit/5fb00becdadfad7c57b5b63c9f8bf6a0a5f2f769#diff-02e6c23bd67643baf08ddbf4d7ec3f5e split the `api.html` page but the reference here was not updated.


stubailo:

PR #305, opened 2017-10-20
expecially -> especially

r-walsh:
Just a typo in the "using-with-types" page.


stubailo:

PR #306, opened 2017-10-25
Fix examples in DOCs to match with release 2.x

jaroslav-kubicek:
As the 2.x, docs contain misleading info how to import `ApolloClient`.


lorensr:
Also `graphql` is a peer dep of client & link

stubailo:

